PR #42, opened 2025-04-18
feat: Output standard out like Hayabusa

### LINKED ISSUE #39: Output standard out like Hayabusa
When outputting to standard output, I would like to output in the same format as Hayabusa:
1. spaces between events
2. alert color for timestamp
3. sigma title and sigma alert level
4. white text for everything else
5. delimiter of orange ` · `
6. If the value is blank we should output ` - `

Note: we haven't supported dynamic fields yet so we don't need to output the field names in orange and values in blue at this time.

@fukusuket Could I ask you to do this one?

fukusuket:
## What Changed
- Closed #39

## Evidence
スクリーンショット 2025-04-18 14 41 17

### DIFF
--- a/src/aws_detect.rs
+++ b/src/aws_detect.rs
@@ -59,7 +59,7 @@ fn write_record(
     csv_writer: &mut Option<Writer<Box<dyn Write>>>,
     json_writer: &mut Option<BufWriter<Box<dyn Write>>>,
     jsonl_writer: &mut Option<BufWriter<Box<dyn Write>>>,
-    std_writer: &mut Option<Writer<Box<dyn Write>>>,
+    std_writer: &mut Option<BufferWriter>,
 ) {
     let record: Vec<String> = profile
         .iter()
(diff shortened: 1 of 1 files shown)


YamatoSecurity:
@fukusuket Thanks so much for this! Very sorry, after looking at it I think it would look better if all the fields were the same color according to the alert with just the ` · ` separator being the orange color. Sorry for the trouble, but is it easy to turn the white text into the same color? 

Also, `-K, --no-color` seems to be working for just the logo, not `Start time`, alerts, etc... could you check this?

fukusuket:
@YamatoSecurity 
Thank you for checking! I updated it!
スクリーンショット 2025-04-18 21 17 27

fukusuket:
### -K 
スクリーンショット 2025-04-19 7 17 12

YamatoSecurity (approved):
@fukusuket LGTM! Thanks so much!

PR #59, opened 2025-04-23
chg: replace sigma-rust to sigmars

### LINKED ISSUE #52: Replace `sigma-rust` crate with `sigmars`?
Many detections for cloud logs will rely on correlation rules. Also, the current `sigma-rust` crate seems to only support sigma v1 field modifiers so we should investigate whether we can use https://crates.io/crates/sigmars/0.2.0 instead which is supposed to support all sigma v2 modifiers as well as correlations.

fukusuket:
## What Changed
- Closed  #52

## Evidence
TBD

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


fukusuket:
@YamatoSecurity 
I’ve quickly replaced it with sigmars for now (though I haven’t tested it yet...😅)!
Could you check if the Correlation rules are working properly? 🙏

fukusuket:
If it seems to be detecting properly, adding multi-threading support should be relatively easy :)

fukusuket:
@YamatoSecurity Sorry! I wasn't able to test it properly.. :( It causes a panic, so I'm currently investigating it.

fukusuket:
Unfortunately, at this point, the correlation rule is not being detected as expected. I’ll raise a question on the sigmars side.

YamatoSecurity:
Thanks for looking into this! Even if correlations is not working at the moment, if this crate detects the same amount as the current one and more with support for v2 modifiers, then we might as well convert over to using this one and work with the authors of sigmars to later get correlations working. What do you think?

fukusuket:
@YamatoSecurity 
After doing a quick comparison, it seems that sigma-rust is currently more stable, so it might be better to replace it with sigmars once it becomes more stable! Also, It might be a good idea to submit issues or pull requests to both crate :)

- Elasped Time
  - main: 31sec
  - This PR(without correlation) 48sec
  - This PR(with correlation) 1m 30sec

- Logic
  - main
      - Probably no issues within the scope of the current rule.
  - This PR   
      - not logic? not working https://github.com/Yamato-Security/suzaku-rules/blob/main/suzaku/aws_cloudtrail_api_created.yml#L26 
      - re modifier? not working https://github.com/Yamato-Security/suzaku-rules/blob/main/sigma/cloud/aws/cloudtrail/aws_cloudtrail_imds_malicious_usage.yml#L23

PR #81, opened 2025-05-26
chg: output one user type

### LINKED ISSUE #79: Output just one user type in `aws-ct-summary`
Right now we print out the total user types like this:
```
Total user types: 1
915,834 - IAMUser (2019-08-23 06:00:07 ~ 2019-08-23 06:00:07)
```

However, since we are filtering by user ARN, there will only be one user type.

Since we have the total event count and first and last timestamp in other columns, I would like to output just the single user type in this field.

Example: `IAMUser`

@fukusuket Can I ask you to do this one?

fukusuket:
## What Changed

- Closed #79 

## Evidence

スクリーンショット 2025-05-26 9 33 25


YamatoSecurity (approved):
@fukusuket LGTM! Thanks so much!

PR #86, opened 2025-05-31
chg: Replace sigma-rust crate with sigmars

### LINKED ISSUE #52: Replace `sigma-rust` crate with `sigmars`?

fukusuket:
## What Changed

- Closed #52

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


fukusuket:
Due to the following issue, the conditions for the list of modifiers are not currently detected correctly.
- https://github.com/crowdalert/sigmars/issues/11

PR #89, opened 2025-06-09
update github actions OSes

YamatoSecurity:
I updated the github actions to use the latest OSes. However, I am not sure if this will cause any problems. Could you please check this?


fukusuket (approved):
I applied the same fix as Hayabusa! LGTM!!🚀
https://github.com/Yamato-Security/suzaku/actions/runs/15561310870

Issue #101, opened 2025-08-02
Support raw JSON output to terminal


YamatoSecurity:
@fukusuket Could I ask you to do this issue next? No hurry but this would be nice as it would make rule testing easier. Right now I have to save the raw JSON data to a file and then `cat` it out so I would like to just be able to output to the terminal.

For example: `./target/release/suzaku aws-ct-timeline -d ../suzaku-sample-data -r test-rule.yml -R`

Right now it gives the error: `--raw-output option is only available in JSON formats. Please specify an output type of 2-5.`

So when `-R` is enabled and `-o` is not specified, I want to output the original JSON log to standard output.

If no JSON output type is specified and `-o` is specified, then we should give this error message.

PR #106, opened 2025-08-07
Finalize 1.0.1

YamatoSecurity:
finalized 1.0.1


YamatoSecurity:
@fukusuket Hummm. could you check why this is failing?

fukusuket:
I added #101 feature!

YamatoSecurity:
@fukusuket Thanks so much! One little thing, can you change it so it will print to terminal without having to specify `-t` ? 

Ex: `./target/release/suzaku aws-ct-timeline -d ../suzaku-sample-data -R`

I think `-t` should only be needed when `-o` is specified.

YamatoSecurity:
@fukusuket Sorry, I think clippy warnings got stricter with the new version of Rust maybe? Can you check this? and then I will release as version 1.1.0 "Obon Release"

fukusuket:
@YamatoSecurity 
An Obon release sounds lovely🍆笑!!  I fixed it!!

PR #109, opened 2025-09-07
feat: implement Azure timeline command and refactor log source handling

### LINKED ISSUE #107: `azure-timeline` command
Create a `azure-timeline` command to support scanning azure cloud logs.

For now we should just work on supporting Activity logs and then Audit logs as they have the most sigma rules.
Signin and risk detection after that.

```
General Options:
  -r, --rules <DIR/FILE>  Specify a custom rule directory or file (default: ./rules)
  -h, --help              Show the help menu

Input:
  -d, --directory <DIR>  Directory of multiple json files
  -f, --file <FILE>      File path to one json file

Output:
  -C, --clobber                    Overwrite files when saving
  -m, --min-level <LEVEL>          Minimum level for rules to load (default: informational)
  -o, --output <FILE>              Save the results to a file
  -t, --output-type <OUTPUT_TYPE>  Output type 1: CSV (default), 2: JSON, 3: JSONL, 4: CSV & JSON, 5: CSV & JSONL [default: 1]
  -R, --raw-output                 Output the original JSON logs (only available in JSON formats or stdout)
      --threads <THREAD NUMBER>    Number of threads to use (default: same as CPU cores)

Display Settings:
  -K, --no-color               Disable color output
  -N, --no-summary             Do not display results summary
  -T, --no-frequency-timeline  Disable event frequency timeline (terminal needs to support Unicode)
  -q, --quiet                  Quiet mode: do not display the launch banner
```

You can reduce the number of options at first to make it easier to implement.

Note: we need to check how the fields differ between activity, audit, etc.. logs as CSV requires common fields that exist in all the logs. For now, we could just support JSON/L and output the raw logs until we figure it out.

Then it would be:
```
General Options:
  -r, --rules <DIR/FILE>  Specify a custom rule directory or file (default: ./rules)
  -h, --help              Show the help menu

Input:
  -d, --directory <DIR>  Directory of multiple json files
  -f, --file <FILE>      File path to one json file

Output:
  -C, --clobber                    Overwrite files when saving
  -m, --min-level <LEVEL>          Minimum level for rules to load (default: informational)
  -o, --output <FILE>              Save the results to a file
  -t, --output-type <OUTPUT_TYPE>  Output type 1: JSONL (default), 2: JSON [default: 1]
      --threads <THREAD NUMBER>    Number of threads to use (default: same as CPU cores)

Display Settings:
  -K, --no-color               Disable color output
  -N, --no-summary             Do not display results summary
  -T, --no-frequency-timeline  Disable event frequency timeline (terminal needs to support Unicode)
  -q, --quiet                  Quiet mode: do not display the launch banner
```

(There probably isn't a need to support outputting JSON and JSONL at the same time...)

fukusuket:
## What Changed

- Closed #107


fukusuket:
@YamatoSecurity 
I’ve completed the initial implementation of azure-timeline!
However, there seem to be some discrepancies between Sigma’s Azure rules and the sample logs I have, which cause many of the rules not to match:

* Case sensitivity mismatches in field names
* Rules referencing non-existent fields

I’ll investigate further to determine whether the issue lies with the Sigma rules or the sample logs.
With the current implementation, We can test using my own rules and the sample logs, so I’d appreciate any comments if you notice anything🙏

fukusuket:
@YamatoSecurity 
After doing some investigation on Azure logs, 
Unfortunately, it seems that the format unfortunately varies depending on the output method...😭:
- Sample logs → JSONL format, not JSON array
- Downloaded directly from the Audit Log portal → JSON array format (not JSONL)
- Downloaded from Log Analytics → the values of the category field seem different?
- When querying in Log Analytics using KQL → field names start with **uppercase** [(e.g. Sentinel query repo)](https://github.com/Azure/Azure-Sentinel/blob/master/Detections/AuditLogs/ChangestoApplicationOwnership.yaml), whereas in the Sample logs they start with **lowercase** → so rules in the Sigma repository seem correct if they are intended for KQL

Therefore, it seems necessary to investigate the specifications of the Azure log JSON format and then decide which format suzaku will support🤔

YamatoSecurity (approved):
@fukusuket LGTM!! Thanks so much for this! I will do some more testing with it.

PR #129, opened 2026-07-08
feat: support M365 Unified Audit Log (CSV + JSON) in azure-timeline

### LINKED ISSUE #128: azure-timeline cannot parse the M365 Unified Audit Log det-eng samples (0 events, CSV skipped)
## Summary

`azure-timeline` cannot detect anything in the official M365 detection-engineering samples at [`suzaku-sample-data/azure/det-eng-samples`](https://github.com/Yamato-Security/suzaku-sample-data/tree/main/azure/det-eng-samples). Running it over all 39 files (19 CSV + 20 JSON) currently reports **0 events / 0 detections**.

## Repro

```
suzaku azure-timeline -d det-eng-samples -o out.csv
# Total log files: 20        <- the 19 .csv files are not counted/read
# Events with hits / Total events: 0 / 0
```

## Root causes (all M365 Unified Audit Log specific)

These samples are **M365 Unified Audit Log** records (`Operation`, `Workload`, `RecordType`, `CreationTime`, …), exported either as JSON or as `Search-UnifiedAuditLog` CSV (real record in the `AuditData` column). Four things block them:

1. **CSV is never read.** `count_files_recursive`, `process_events_from_dir`, and `get_content` only accept `.json`/`.gz`, so all 19 CSV files are silently skipped.
2. **All events are dropped at the time filter.** `filter_by_time` returns `false` when the profile timestamp key is absent — and it always is for UAL (which uses `CreationTime`, not `.time`/`.eventTimestamp`). So even with **no** `--timeline-start/-end/-offset`, every UAL event is filtered out → `Total events: 0`.
3. **Single / pretty-printed JSON objects aren't parsed.** `log_contents_to_events` / `load_json_from_file` handle a JSON array or a `{ "value": [...] }` object, but a bare single record object (common in these samples, some pretty-printed) falls through to the JSONL path and yields nothing.
4. **`is_match_service` rejects UAL.** The `activitylogs`/`auditlogs`/`signinlogs` services all key on the Azure Monitor `category` field, which UAL records don't have, so no rule can ever fire on a UAL event.

## Proposed fix

Add M365 Unified Audit Log support to `azure-timeline`:
- Read `.csv` exports and unwrap the `AuditData` column; normalize the `AuditData` wrapper (string in CSV, object in JSON) and fold the UAL Name/Value property bags (`ExtendedProperties`/`DeviceProperties`/`Parameters`/`ModifiedProperties`) into objects so rules can match nested values like `ExtendedProperties.UserAgent`.
- Parse a bare/pretty single JSON record object.
- Don't drop events at `filter_by_time` when no time filter is set; parse the timezone-less `CreationTime` (assume UTC) when one is.
- Add an `m365` service to the Azure log source that recognizes UAL records (by `Workload`/`RecordType`), and map `CreationTime` in `azure_profile.yaml`.

PR to follow.

Shirofune-Security:
Closes #128.

## Problem

`azure-timeline` could not detect anything in the official M365 samples at [`suzaku-sample-data/azure/det-eng-samples`](https://github.com/Yamato-Security/suzaku-sample-data/tree/main/azure/det-eng-samples) — 39 files (19 CSV + 20 JSON) produced **0 events / 0 detections**. These are M365 Unified Audit Log records (`Operation`/`Workload`/`RecordType`/`CreationTime`), exported as JSON or as `Search-UnifiedAuditLog` CSV (real record in the `AuditData` column). See #128 for the four root causes.

## Changes

- **CSV support** — `count_files_recursive`, `process_events_from_dir`, and `get_content` now accept `.csv`; rows are parsed with the `csv` crate (already a dependency).
- **Record normalization** (`normalize_azure_event`) — unwraps the `AuditData` wrapper (a JSON *string* in CSV exports, a nested *object* in JSON exports), then folds the UAL Name/Value property bags (`ExtendedProperties`/`DeviceProperties`/`Parameters`/`ModifiedProperties`) into objects so rules can match nested values like `ExtendedProperties.UserAgent`. Gated to UAL records (by `Workload`/`RecordType`); Azure Monitor diagnostic events are untouched.
- **Single-object JSON** — `log_contents_to_events` / `load_json_from_file` now parse a bare or pretty-printed single record object (previously only arrays and `{ "value": [...] }` were handled).
- **Time filter** — `filter_by_time` returns `true` when no `--timeline-start/-end/-offset` is set (instead of dropping events whose timestamp key is absent), and parses the timezone-less UAL `CreationTime` (assumed UTC) when a filter is set.
- **Service matching** — added an `m365` service to the Azure log source that recognizes UAL records (by `Workload`/`RecordType`); mapped `CreationTime` in `azure_profile.yaml`.

## Verification

- `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings` clean.
- **70 unit tests pass** (8 new: `AuditData` string/object unwrap, Name/Value folding, non-UAL passthrough, CSV parse, single-object JSON, no-filter passthrough, `CreationTime` naive parse).
- End to end over the 39 det-eng samples with a companion set of 21 M365 Sigma rules: **39 files read, 125 events parsed** (was 0), **every one of the 39 files produces ≥1 detection** (111/125 events hit).

The 21 M365 Sigma rules used to validate this are a separate deliverable (intended for `suzaku-rules`).


Shirofune-Security:
Follow-up in this PR: reworked the **Azure output profile** so M365 UAL detections actually carry DFIR context. Previously every UAL row was `Category,OperationName,Entity,Caller,SrcIP,ResourceId,CorrelationId` = all `-` (those are Azure-Monitor field names). The profile now extracts the UAL fields (via `|` alternation, so Azure Monitor diagnostic logs still populate):

`Timestamp, Workload, Operation, RecordType, Result, User, SrcIP, TargetObject, UserAgent, AppId, LogonError, Details, ...`

`Details` summarizes the actual change — the Exchange cmdlet `Parameters` or directory `ModifiedProperties` — e.g.:

- `Set-Mailbox` → `Details: DeliverToMailboxAndForward: True, ForwardingSmtpAddress: johndoe@gmail.com` (the exfil target)
- `Add member to role.` → `Details: Role.DisplayName: Global Administrator`
- `Set-AdminAuditLogConfig` → `Details: UnifiedAuditLogIngestionEnabled: False`
- AzureHound sign-in → `UserAgent: azurehound/v2.0.4`

`Details` is synthesized as a deterministic string in `normalize_azure_event` (rendering sigma_rust's `HashMap`-backed event value directly would vary across runs). Verified deterministic across repeated runs, populated identically from both CSV and JSON, and detection coverage is unchanged (39/39 samples).

YamatoSecurity (approved):

PR #134, opened 2026-07-10
fix: remove PDF generation steps from release workflow

fukusuket:
## What Changed
Removed the PDF generation step from GitHub release actions.

## Evidence
After 
- https://github.com/Yamato-Security/suzaku/actions/runs/29128719475

Before
- https://github.com/Yamato-Security/suzaku/actions/runs/29101268272


YamatoSecurity (approved):
@fukusuket Thanks! I forgot to remove the PDF creation now that we don't need it.

PR #136, opened 2026-07-12
Add ignore-list support to skip loading rules by UUID

### LINKED ISSUE #135: Support an ignore-list file to skip loading rules by UUID
## Summary

Add support for an **ignore-list file** so that specific rules can be excluded from loading by their `id` (UUIDv4), without deleting the rule files.

## Motivation

Some rules become redundant — e.g. a plain `eventName`-only rule that is superseded by a richer rule for the same event. We want to keep the original `.yml` in [suzaku-rules](https://github.com/Yamato-Security/suzaku-rules) (for history/reference) but **not load it**, to avoid duplicate detections.

The suzaku-rules repo already carries such a list at `config/aws_ignore_rule_list.txt` (one UUID per line), but Suzaku does not consume it yet.

## Proposed behavior

- When running `aws-ct-timeline` (and, by extension, other timeline commands), read `config/<logsource>_ignore_rule_list.txt` from the rules directory (e.g. `./rules/config/aws_ignore_rule_list.txt`).
- Any loaded rule whose `id` matches a UUID in that file is **skipped** (not loaded, not evaluated).
- File format (same as the existing `filtered_sigma_rules.txt`): one UUID per line; blank lines and lines starting with `#` are ignored; an inline `# comment` after the UUID is allowed.
- Print how many rules were skipped, for transparency.
- No-op when the file is absent.

## Notes

- Azure would use `azure_ignore_rule_list.txt` under the same convention.

Shirofune-Security:
Closes #135.

## Summary

Adds support for an **ignore-list file** so specific rules can be excluded from loading by their `id` (UUID), without deleting the rule files. This lets superseded/duplicate rules stay in [suzaku-rules](https://github.com/Yamato-Security/suzaku-rules) for reference while not being loaded (avoiding duplicate detections).

## Behavior

- On timeline load, reads `<rules-dir>/config/<log>_ignore_rule_list.txt` — for AWS that's `./rules/config/aws_ignore_rule_list.txt` (Azure would use `azure_ignore_rule_list.txt`).
- Any loaded rule whose `id` matches a UUID in the file is **skipped**.
- File format (same as the existing `filtered_sigma_rules.txt`): one UUID per line; blank lines and lines starting with `#` are ignored; an inline `# comment` after the UUID is allowed.
- Prints `Rules skipped via ignore-list: N`. No-op when the file is absent.

## Implementation

- `src/core/rules.rs`: `ignore_rule_list_path()`, `load_ignore_rule_ids()`, `filter_ignored_rules()` (+ unit tests).
- `src/core/log_source.rs`: `ignore_rule_list_filename()` per log source.
- `src/core/timeline.rs`: filter loaded rules through the ignore-list before the level filter; report the skipped count.

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` all pass (5 new unit tests).
- End-to-end with `aws-ct-timeline` against a rules dir containing `config/aws_ignore_rule_list.txt` (16 UUIDs): output shows `Rules skipped via ignore-list: 16`, and the listed rules no longer appear in the timeline — verified a superseded plain rule (`2070cb71-…`, *IAM Login Profile Created*) is gone while the replacement rules still fire.

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -6,6 +6,7 @@
 
 - Azureログ用のDFIRタイムラインを作成する`azure-timeline`コマンドを追加した。 (#109) (@fukusuket)
 - CloudTrailログを検索するための`aws-ct-search`コマンドを追加した。(#117) (@fukusuket)
+- UUIDを指定してルールを読み込み対象から除外できる除外リストファイル（`config/aws_ignore_rule_list.txt`）に対応した。これにより、置き換えられた重複ルールをリポジトリに残したまま読み込まないようにできる。 (#136) (@YamatoSecurity)
 
 **改善:**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Added the `azure-timeline` command to create a DFIR timeline for Azure logs. (#109) (@fukusuket)
 - New `aws-ct-search` command to search through CloudTrail logs. (#117) (@fukusuket)
+- Added support for an ignore-list file (`config/aws_ignore_rule_list.txt`) to skip loading rules by UUID, so superseded/duplicate rules can stay in the repo without being loaded. (#136) (@YamatoSecurity)
 
 **Enhancements:**
 
--- a/src/core/log_source.rs
+++ b/src/core/log_source.rs
@@ -15,6 +15,15 @@ impl LogSource {
         }
     }
 
+    /// File name (under the rules directory's `config/`) listing rule UUIDs to skip loading.
+    pub fn ignore_rule_list_filename(&self) -> &str {
+        match self {
+            LogSource::Aws => "aws_ignore_rule_list.txt",
+            LogSource::Azure => "azure_ignore_rule_list.txt",
+            LogSource::All => "",
+        }
+    }
+
     pub fn supported_services(&self) -> &[&str] {
         match self {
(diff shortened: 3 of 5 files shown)


Shirofune-Security:
Good catch — fixed in e73d4cc.

The ignore-list is now applied to correlation and base rules as well, not just normal detection rules:
- **Base rules** whose `id` is ignore-listed are no longer registered into the `CorrelationEngine`. Because a correlation is only added when `added_base_rules == total_base_rules`, skipping one of its base rules also cascade-skips the dependent correlation (you can't run a correlation whose base rule was explicitly ignored).
- **Correlation rules** whose own `id` is ignore-listed are no longer registered.
- Both are now included in the `Rules skipped via ignore-list: N` count.

Refactored the check into a shared `rules::is_ignored()` used by `filter_ignored_rules` and the correlation loop, with a unit test.

Verified end-to-end with the `Many Recon Events` correlation (`08d8b077…`) and its base rule `Recon Event` (`33ecb0d0…`): with no ignore-list the run reports `Total correlation rules: 1`; adding either the correlation UUID or the base-rule UUID to `config/aws_ignore_rule_list.txt` drops it to `Total correlation rules: 0` with `Rules skipped via ignore-list: 1` — so it's removed from the engine and cannot fire. `cargo fmt`/`clippy -D warnings`/`cargo test` (85 tests) all pass.

YamatoSecurity (approved):

PR #142, opened 2026-07-17
Support Sigma `tags` field as an abbreviated Tags column

### LINKED ISSUE #62: Support sigma `tags` field
The `tags` field in sigma rules is a list so cannot currently be outputted.
It would be great if Suzaku could automatically detect if the field value is a list and if so concatenate the values together with a separator but if it is hard to detect automatically then manually support certain field names like `tags`.

* Multiple tags should be joined together with ` ¦ ` like Hayabusa does.
* The following abbreviations should be done:
1. `attack.reconnaissance` -> `Recon`
2. `attack.resource-development` -> `ResDev`
3. `attack.initial-access` -> `InitAccess`
4. `attack.execution` -> `Exec`
5. `attack.persistence` -> `Persis`
6. `attack.privilege-escalation` -> `PrivEsc`
7. `attack.defense-evasion` -> `Evas`
8. `attack.credential-access` -> `CredAccess`
9. `attack.discovery` -> `Disc`
10. `attack.lateral-movement` -> `LatMov`
11. `attack.collection` -> `Collect`
12. `attack.command_and_control` -> `C2`
13. `attack.exfiltration` -> `Exfil`
14. `attack.impact` -> `Impact`

* `attack.t` should be converted to just `T` so that `attack.t1562.001` converts to `T1562.001` 
* `attack.g` should be converted to just `G` so that `attack.g0035` converts to `G0035`

Example:
```
tags:
    - attack.g0035
    - attack.credential_access
    - attack.discovery
    - attack.t1110
    - attack.t1087
```

will convert to `G0035 ¦ CredAccess ¦ Disc ¦ T1110 ¦ T1087`

For JSON output as well, instead of creating an array, lets keep it a flat JSON file for now with the same string as for CSV output.

Shirofune-Security:
## Summary

Closes #62.

Sigma rules carry a `tags` list (ATT&CK tactics/techniques/groups) that suzaku previously could not output. This adds a **`Tags` column** to both the `aws-ct-timeline` and `azure-timeline` output that renders the list as a single ` ¦ `-separated string (like Hayabusa) with ATT&CK abbreviations.

### Behavior

Example:

```yaml
tags:
  - attack.g0035
  - attack.credential_access
  - attack.discovery
  - attack.t1110
  - attack.t1087
```

renders as: `G0035 ¦ CredAccess ¦ Disc ¦ T1110 ¦ T1087`

- **ATT&CK tactics** are abbreviated via a **config table**, `config/mitre_tactics.txt` (see below).
- **Techniques** `attack.t1562.001` → `T1562.001`; **groups** `attack.g0035` → `G0035` (algorithmic, in code).
- Both the **hyphen and underscore** tactic spellings are folded to the same abbreviation — both forms appear in the loaded rule corpus.
- Unknown tag namespaces (e.g. `cve.*`, `car.*`) pass through unchanged.
- **JSON/JSONL** output keeps the value as a **flat string** (not an array).

### Tactic table is not hard-coded

The tactic short names live in `config/mitre_tactics.txt` — the **same table Hayabusa ships**, minus its `html_tag_output_str` column. It is a simple `<full tag>,<abbreviation>` CSV, loaded once at runtime into a process-wide map (same runtime-config pattern as the output profiles). Users can edit it, and it stays in sync with Hayabusa.

> ⚠️ Because the table now comes from Hayabusa, `attack.defense-evasion` maps to **`Stealth`** (not `Evas` as originally listed in #62), and `attack.stealth` / `attack.defense-impairment` are also recognized. Output follows Hayabusa's table.

### Implementation

- `config/mitre_tactics.txt`: new tactic-abbreviation table (Hayabusa's, without the HTML column).
- `config/aws_profile.yaml` / `config/azure_profile.yaml`: new `Tags: 'sigma.tags'` column (before `RuleID`).
- `src/core/timeline_writer.rs`: `load_mitre_tactics()` + a `OnceLock` cache; `abbreviate_tag()` / `format_tags()`; `RuleInfo::tags()` now joins with ` ¦ ` and abbreviates instead of `tags.join(", ")`. Covers normal and correlation rules. A missing table degrades gracefully (tactics pass through; techniques/groups unaffected).

### Testing

- Unit tests: all tactic mappings (loaded from the real `config/mitre_tactics.txt`), the loader (header skip, extra-column tolerance, missing-file fallback), hyphen/underscore normalization, techniques/groups, mixed-case, unknown-namespace passthrough, and the issue example.
- `cargo build`, `clippy --all-targets -- -D warnings`, `fmt -- --check`, and `cargo test` (102 tests) all pass.
- Verified end-to-end against `test_files/json`: CSV `Stealth ¦ T1562.008`, `PrivEsc ¦ Stealth ¦ InitAccess ¦ Persis ¦ T1078.004`; JSON emits `"Tags"` as a flat string.


Shirofune-Security:
Updated per review feedback: the MITRE ATT&CK tactic abbreviations are no longer hard-coded. They now load at runtime from a new `config/mitre_tactics.txt` — the same table Hayabusa ships, minus the `html_tag_output_str` column. Two behavior notes that follow from using Hayabusa's table: `attack.defense-evasion` now abbreviates to **Stealth** (not `Evas`), and `attack.stealth`/`attack.defense-impairment` are recognized too. Technique/group shortening stays in code since it's algorithmic, not a lookup.

YamatoSecurity (approved):

Shirofune-Security:
Added a docs page: **Rules → MITRE ATT&CK Tags** (`website/docs/rules/mitre-attack-tags.md` + Japanese `.ja.md`, plus the nav entry in `mkdocs.yml`). It explains the `Tags` column, the ` ¦ ` join, and how tactics (via `config/mitre_tactics.txt`), techniques and groups are abbreviated — including that `attack.defense-evasion`/`attack.stealth` both map to `Stealth` following the MITRE ATT&CK v19 (April 2026) rename of Defense Evasion → Stealth. Verified with `mkdocs build --strict` locally (the docs workflow only runs on merge to main, not on PRs).

PR #152, opened 2026-07-18
Replace input/filesystem unwraps with clean errors (#149)

### LINKED ISSUE #149: Panics instead of clean errors on filesystem edge cases (non-UTF8 filenames, walk errors, unwritable output paths)
Suzaku is a locally-run forensic CLI that scans directories of cloud logs. A few input/filesystem-facing spots use `unwrap()`/`expect()` on `Result`/`Option` values that are genuinely fallible in the field — a non-UTF-8 filename in the input tree, a directory-walk/permission error, or an unwritable `--output` path. Each of these aborts the whole run with a Rust panic and backtrace instead of a clean diagnostic, and in the first two cases a single odd file anywhere under the scanned directory kills the entire scan before any results are produced. This issue proposes a small hardening batch that replaces these input/FS `unwrap`s with clean, actionable errors.

### 1. Non-UTF-8 filename panics the file count (`src/core/scan.rs:587`)

`count_files_recursive` guards only that the *extension* is valid UTF-8 (line 584), then unconditionally unwraps the *whole path* to a `&str` on line 587:

```rust
if path.is_file() {
    if let Some(ext) = path.extension().and_then(|s| s.to_str())
        && (ext == "json" || ext == "jsonl" || ext == "gz" || ext == "csv")
    {
        let path_str = path.to_str().unwrap();
        if !filter_file_by_date_path(file_date_opt, path_str) {
            continue;
        }
        count += 1;
        total_size += fs::metadata(&path)?.len();
        paths.push(path_str.to_string());
    }
} else if path.is_dir() {
    let (sub_count, sub_paths, sub_size) = count_files_recursive(&path, file_date_opt)?;
    // ...
}
```

On Linux, filenames are arbitrary byte sequences and need not be valid UTF-8. A file whose *extension* is valid ASCII (e.g. `.json`) but whose stem or a parent directory component is not UTF-8 passes the `ext` guard and reaches `path.to_str().unwrap()`, which returns `None` and panics. Because `count_files_recursive` is called at the very start of `process_events_from_dir` (`src/core/scan.rs:125`) — before any log file is processed — one oddly-named file anywhere in the tree aborts the entire scan.

### 2. Directory-walk error panics the scan (`src/core/scan.rs:90`)

`scan_directory` (returns `()`) calls the fallible `process_events_from_dir` and unwraps its `Result`:

```rust
    process_events_from_dir(
        process_events,
        d,
        options.output_opt.output.is_some(),
        no_color,
        log,
        &options.input_opt.file_date_opt,
    )
    .unwrap();
```

`process_events_from_dir` returns `Result<(), Box<dyn Error>>` and propagates `fs::read_dir`/`entry?`/`fs::metadata` errors from the recursive walk. A `read`-denied subdirectory (mode `000`), or a file that is removed between the `path.is_file()` check and the later `fs::metadata` call, turns into `Err(..)` that this `.unwrap()` converts into a panic with a backtrace rather than a readable message.

### 3. Unwritable `--output` path panics before writing (`src/core/util.rs:14`, `:23`)

Both writer constructors unwrap `File::create` on the user-supplied output path:

```rust
pub fn get_writer(output: &Option<PathBuf>) -> Writer<Box<dyn Write>> {
    let wtr: Writer<Box<dyn io::Write>> = if let Some(output) = output {
        Writer::from_writer(Box::new(fs::File::create(output).unwrap()))
    } else {
        Writer::from_writer(Box::new(io::stdout()))
    };
    wtr
}

pub fn get_json_writer(output: &Option<PathBuf>) -> BufWriter<Box<dyn Write>> {
    if let Some(output) = output {
        let file = File::create(output).expect("Failed to create file");
        BufWriter::new(Box::new(file))
    } else {
        BufWriter::new(Box::new(std::io::stdout()))
    }
}
```

A `--output` pointing at a non-existent directory, a read-only location, or a path the user lacks permission to create panics (with a backtrace, or the terse `Failed to create file` from the `.expect`) instead of a clean "cannot write to X" message.

## Impact

- **Who:** analysts running suzaku over collected cloud-log corpora (AWS CloudTrail JSON/JSONL/gz, Azure/M365 exports) and anyone scripting suzaku in a pipeline.
- **Severity:** robustness / clean-shutdown, not memory safety — no RCE. The failure mode is an unexpected panic with a backtrace and a non-zero abort, which is a poor UX for a forensic tool and, for cases (1) and (2), a denial of the entire scan triggered by attacker-influenceable input: the log files being triaged often originate from an attacker-touched environment, and a directory containing one non-UTF-8 filename (or one unreadable subdir) is enough to abort triage of an otherwise-good corpus.
- **Concrete triggers / repro:**
  - (1) `touch "$(printf 'evil-\xff.json')"` in the scan directory, then run any `suzaku` timeline/scan over that directory → panic at `scan.rs:587` before any file is read.
  - (2) `mkdir sub && chmod 000 sub` inside the scan directory (as a non-root user) → panic at `scan.rs:90` via the propagated `read_dir` error.
  - (3) `suzaku ... --output /nonexistent-dir/out.csv` → panic at `util.rs:14`.

## Suggested fix

Replace these input/FS `unwrap`/`expect` calls with lossy conversion or propagated errors and clean diagnostics.

**`src/core/scan.rs:587`** — use `to_string_lossy()`; the downstream uses are a date-prefix filter and display/collection, so a lossy rendering is acceptable and never panics:

```rust
                let path_str = path.to_string_lossy();
                if !filter_file_by_date_path(file_date_opt, &path_str) {
                    continue;
                }
                count += 1;
                total_size += fs::metadata(&path)?.len();
                paths.push(path_str.into_owned());
```

(`filter_file_by_date_path` takes `&str`; `&path_str` deref-coerces from `&Cow<str>`. `into_owned()` yields the `String` for `paths`.)

**`src/core/scan.rs:90`** — `scan_directory` returns `()`, so handle the error in place (log and continue) instead of unwrapping:

```rust
    if let Err(e) = process_events_from_dir(
        process_events,
        d,
        options.output_opt.output.is_some(),
        no_color,
        log,
        &options.input_opt.file_date_opt,
    ) {
        AlertMessage::alert(&format!("Failed to scan directory {}: {e}", d.display())).ok();
    }
```

(Use whatever the crate's standard error-print helper is; if there is none in scope, `eprintln!` is an acceptable minimal form. Alternatively, thread a `Result` up through `scan_directory` if callers can handle it.)

**`src/core/util.rs:14` and `:23`** — return `Result` and map the IO error to a clean message. This is a shared helper, so the change belongs here and callers should be updated to handle the `Result` (e.g. print the message and exit non-zero):

```rust
use std::error::Error;

pub fn get_writer(output: &Option<PathBuf>) -> Result<Writer<Box<dyn Write>>, Box<dyn Error>> {
    let wtr: Writer<Box<dyn io::Write>> = if let Some(output) = output {
        let file = fs::File::create(output)
            .map_err(|e| format!("Cannot write to output file {}: {e}", output.display()))?;
        Writer::from_writer(Box::new(file))
    } else {
        Writer::from_writer(Box::new(io::stdout()))
    };
    Ok(wtr)
}

pub fn get_json_writer(output: &Option<PathBuf>) -> Result<BufWriter<Box<dyn Write>>, Box<dyn Error>> {
    if let Some(output) = output {
        let file = File::create(output)
            .map_err(|e| format!("Cannot write to output file {}: {e}", output.display()))?;
        Ok(BufWriter::new(Box::new(file)))
    } else {
        Ok(BufWriter::new(Box::new(std::io::stdout())))
    }
}
```

### Checklist

- [ ] `src/core/scan.rs:587` — `path.to_str().unwrap()` → `to_string_lossy()`
- [ ] `src/core/scan.rs:90` — `process_events_from_dir(..).unwrap()` → log-and-continue (or propagate)
- [ ] `src/core/util.rs:14` — `get_writer` returns `Result`, clean diagnostic; update callers
- [ ] `src/core/util.rs:23` — `get_json_writer` returns `Result`, clean diagnostic; update callers

## Verification

Add a unit test alongside the existing `#[cfg(test)]` tests in `src/core/scan.rs` for the non-UTF-8 case (Unix-only), asserting the walk no longer panics and still counts the file:

```rust
#[cfg(test)]
mod fs_edge_tests {
    use super::*;
    #[cfg(unix)]
    #[test]
    fn count_files_recursive_handles_non_utf8_filename() {
        use std::ffi::OsStr;
        use std::os::unix::ffi::OsStrExt;

        let dir = tempfile::tempdir().unwrap();
        // "bad-\xff.json": valid extension, non-UTF-8 stem.
        let name = OsStr::from_bytes(b"bad-\xff.json");
        let path = dir.path().join(name);
        std::fs::write(&path, b"[]").unwrap();

        let (count, paths, _size) =
            count_files_recursive(&dir.path().to_path_buf(), &FileDateOption::default())
                .expect("should not panic on non-UTF-8 filename");
        assert_eq!(count, 1);
        assert_eq!(paths.len(), 1);
    }
}
```

Manual checks for the other two:

```sh
# (2) walk error is reported, not a panic
mkdir -p corpus/locked && chmod 000 corpus/locked
suzaku <timeline-cmd> -d corpus   # expect a clean "Failed to scan directory ..." line, non-zero exit

# (3) unwritable output is reported, not a panic
suzaku <timeline-cmd> -d corpus --output /nonexistent-dir/out.csv
# expect "Cannot write to output file /nonexistent-dir/out.csv: ..." and no backtrace
```

(Run panicking builds without `RUST_BACKTRACE` set to confirm the before/after difference is a clean message rather than a panic.)

---

Shirofune-Security:
## Summary

Closes #149. Replaces the input/filesystem `unwrap()`/`expect()` calls that aborted a run with a Rust panic + backtrace with clean, actionable errors. Covers **all three** cases in the issue **plus** a follow-on panic that a partial fix would have exposed.

## Changes

**1. Non-UTF-8 filename — `scan.rs`**
- `count_files_recursive`: `path.to_str().unwrap()` → `path.to_string_lossy()`. A file whose extension is valid (`.json`) but whose stem/parent is not UTF-8 no longer panics the count that runs *before* any file is processed.
- `process_events_from_dir` (progress bar): `fs::metadata(&path).unwrap()` → `.map(|m| m.len()).unwrap_or(0)`. **This is the key follow-on gap:** once the count walk stores a *lossy* path for a non-UTF-8 file, that path doesn't resolve, so with `-o`/`--output` (which turns on `show_progress`) the very next `fs::metadata(...).unwrap()` still panicked. Now it doesn't.

**2. Directory-walk errors** — the four `process_events_from_dir(..).unwrap()` sites (`scan_directory`, and the `aws-ct-search` / `aws-ct-metrics` / `aws-ct-summary` commands) now print `Failed to scan directory <path>: <err>` and continue instead of panicking (e.g. an unreadable `chmod 000` subdir, or a file removed mid-scan).

**3. Unwritable `--output` — `util.rs`**
- `get_writer` / `get_json_writer` now return `Result<_, String>` and map the IO error to `Cannot write to output file <path>: <err>`. `init_writers` propagates it; a new `fatal_error(no_color, msg) -> !` helper prints the message and `exit(1)` (non-zero, no backtrace). Callers (`make_timeline`, `aws-ct-search`, `aws-ct-metrics`, `aws-ct-summary`) updated.

## Verification

- New unit tests: non-UTF-8 count walk; **non-UTF-8 walk with `show_progress = true`** (the case a count-only fix misses); and the `get_writer`/`get_json_writer` `Result` error paths. `cargo fmt`, `clippy --all-targets -D warnings`, and `cargo test` (107 tests) all pass.
- End-to-end (with `RUST_BACKTRACE=1`, so a panic would be obvious):
  - non-UTF-8 `.json` + `-o` → completes, exit 0, no panic.
  - `chmod 000` subdir → `Failed to scan directory …: Permission denied (os error 13)`, no panic.
  - `--output /nonexistent-dir/out.csv` → `Cannot write to output file …: No such file or directory (os error 2)`, exit 1, no backtrace.


Shirofune-Security:
## Verification against real sample data (`suzaku-sample-data`)

Built release binaries of `main` and this branch and ran both against the AWS CloudTrail logs in [suzaku-sample-data](https://github.com/Yamato-Security/suzaku-sample-data) (310 files / 596 KiB, 770 rules loaded).

> Note: the sample-data repo currently contains only AWS logs (`aws/traildiscover.cloud/`) — no Azure/M365 set — so Azure was not exercised here. The changed code path (`scan.rs` directory walk + `util.rs` writers) is shared by `azure-timeline`, and its unit tests cover it.

### Results parity (clean data) — no change
| | main | this PR |
|---|---|---|
| Events with hits / total | 295 / 310 | 295 / 310 |
| High detections | 7 | 7 |
| CSV output | — | **byte-identical** to main |
| JSONL output (sorted) | — | **identical** to main |

### Performance (`--debug`, median of 5 runs) — no measurable impact
| | elapsed | user CPU | sys CPU | peak RSS |
|---|---|---|---|---|
| main | 0.074s | 0.125s | 0.065s | 12.6 MiB |
| this PR | 0.074s | 0.112s | 0.075s | 12.7 MiB |

Within run-to-run noise; expected, since the change only swaps `to_str().unwrap()` for `to_string_lossy()` and adds error handling on failure paths.

### Edge-case behavior — the actual fix (`RUST_BACKTRACE=1`, so a panic is loud)
Using a copy of the sample dir with one non-UTF-8 `.json` file added:

| scenario | main | this PR |
|---|---|---|
| non-UTF-8 filename + `-o` | **panic (exit 101)** | ok, exit 0 |
| unwritable `--output` path | **panic (exit 101)** | `Cannot write to output file …`, exit 1 |
| unreadable subdir (`chmod 000`) | **panic (exit 101)** | `Failed to scan directory …`, exit 0 |

main panics in all three; this PR produces a clean message (and a non-zero exit for the fatal output-path case) in every one.

YamatoSecurity (approved):

Shirofune-Security:
Both fixed in the latest commit — good catches.

**1. Non-UTF-8 files were counted but not processed.** You're right; the `to_string_lossy()` rendering didn't resolve back to the real file, so it was skipped at read time. `count_files_recursive` now returns `Vec<PathBuf>` and keeps the **real path** through the pipeline; the lossy string is used only for the date filter, the extension checks, and the progress display. Reproduced your exact scenario with `aws-ct-metrics`:

```
Total log files: 2
| UpdateTrail | 50.00% | 1 |
| DeleteTrail | 50.00% | 1 |   <- the non-UTF-8-named file's event is now processed (was missing before)
```

Added a regression test (`process_events_from_dir_reads_non_utf8_filename`) asserting the event in a `bad-\xff.json` file is actually processed, not skipped.

**2. aws-ct-summary JSON/JSONL still panicked on unwritable output.** The `-t 2/3/4/5` paths used direct `File::create().unwrap()` — I missed them. Both now map the IO error through `fatal_error()`, matching the CSV path:

```
$ suzaku aws-ct-summary -f test_files/json/UpdateTrail.json -o /tmp/missing/out -t 2
Cannot write to output file /tmp/missing/out.json: No such file or directory (os error 2)   (exit 1, no backtrace)
```

`clippy --all-targets -D warnings`, `fmt`, and `cargo test` (120 tests) all pass.

PR #153, opened 2026-07-18
Fix --geo-ip overwriting every column with the raw sourceIPAddress for non-IP values (#145)

### LINKED ISSUE #145: GeoIP enrichment overwrites every output column with the raw sourceIPAddress when it is not a valid IP (e.g. `cloudtrail.amazonaws.com`)
When `--geo-ip` is enabled, any CloudTrail record whose `sourceIPAddress` is not a parseable IP address (a routine occurrence for AWS-service-initiated events, e.g. `cloudtrail.amazonaws.com`, `AWS Internal`, `ecs-tasks.amazonaws.com`, `health.amazonaws.com`) has **every** column of its timeline row — `Timestamp`, `EventName`, `RuleTitle`, `Level`, etc. — silently replaced by that raw `sourceIPAddress` string. The GeoIP block in `get_value_from_event_common` returns the raw IP string for *all* keys (not just the three geo columns) when the address fails to parse. There is no error or warning; the output (CSV/JSON/JSONL/stdout) is simply corrupted for exactly the AWS-service events, which are often the ones an analyst most needs to read.

## Problem

`get_value_from_event_common` (`src/core/timeline_writer.rs:384-448`) is called **once per output column** — `write_record` maps it over every entry in the profile (`src/core/timeline_writer.rs:69-73`):

```rust
let mut record: Vec<String> = context
    .profile
    .iter()
    .map(|(_k, v)| get_value_from_event(v, event, rule, context.geo, localtime))
    .collect();
```

The GeoIP block at the top of the function runs for **every one of those calls** (`src/core/timeline_writer.rs:391-407`):

```rust
    // GeoIP処理部分（共通）
    if let Some(geo) = geo_ip
        && let Some(ip) = event.get("sourceIPAddress")
    {
        let ip = ip.value_to_string();
        if let Some(ip) = geo.convert(ip.as_str()) {
            if key == "SrcASN" {
                return geo.get_asn(ip);
            } else if key == "SrcCity" {
                return geo.get_city(ip);
            } else if key == "SrcCountry" {
                return geo.get_country(ip);
            }
        } else {
            return ip;
        }
    }
```

`geo.convert(...)` (`src/option/geoip.rs:65-75`) returns `Some` only when `IpAddr::from_str` succeeds, and `None` otherwise:

```rust
    pub fn convert(&self, ip: &str) -> Option<IpAddr> {
        let ip = if ip.starts_with("::ffff:") {
            ip.replace("::ffff:", "")
        } else {
            ip.to_string()
        };
        if let Ok(ip) = IpAddr::from_str(&ip) {
            return Some(ip);
        }
        None
    }
```

The bug is the `else` branch at `src/core/timeline_writer.rs:404-406`: when `convert()` returns `None` (i.e. `sourceIPAddress` is not a valid IP), it does `return ip;` — the raw `sourceIPAddress` string — **regardless of which column `key` is being computed**. Because the function is invoked per column, the same string is returned for `Timestamp`, `EventName`, `RuleTitle`, `Level`, and every other column, so the entire row is filled with the service-principal string.

Note that the `Some` (valid-IP) path is already correct: only `SrcASN`/`SrcCity`/`SrcCountry` return geo values and all other keys fall through to normal field processing. The defect is isolated to the `None` branch, which was never scoped to the three geo keys.

## Impact

- **Affected:** anyone running `suzaku aws-ct-timeline` (or any output that flows through `write_record`) **with `--geo-ip` enabled**. Without `--geo-ip` the GeoIP block is skipped entirely and there is no issue.
- **Effect:** silent output/data corruption. Every column of an affected record is overwritten. This is not a crash and produces no diagnostic — a spreadsheet-based analyst sees rows where `Timestamp`, `EventName`, `RuleTitle`, `Level`, etc. all read `cloudtrail.amazonaws.com`, and may not realize the original field values were lost. It corrupts precisely the AWS-service events (control-plane / managed-service activity), which are frequently relevant to an investigation.
- **Trigger is not exotic:** CloudTrail routinely populates `sourceIPAddress` with a service principal rather than an IP for AWS-initiated events. The `sourceIPAddress` field is attacker-influenceable in general, but no crafted input is needed here — normal AWS logs contain these values.
- Not a code-execution or injection issue; the corrupted values are the log's own strings written into the analyst's timeline.

**Concrete repro:** Take any CloudTrail event with a non-IP `sourceIPAddress`, e.g. `{"eventTime":"2023-07-10T12:27:45Z","eventName":"AssumeRole","sourceIPAddress":"cloudtrail.amazonaws.com", ...}`, and run:

```
suzaku aws-ct-timeline <input-dir> -G <maxmind-db-dir>
```

Every column of that record's row in the output is `cloudtrail.amazonaws.com`.

## Suggested fix

Scope the GeoIP block so it only runs for the three geo columns, and on a parse failure (or a missing DB / missing `sourceIPAddress`) return the `-` placeholder **for those keys only** — never `return ip` for an arbitrary column. Replace `src/core/timeline_writer.rs:391-407` with:

```rust
    // GeoIP処理部分（共通）: only the three geo columns are enriched.
    if matches!(key, "SrcASN" | "SrcCity" | "SrcCountry") {
        if let Some(geo) = geo_ip
            && let Some(ip) = event.get("sourceIPAddress")
        {
            let ip = ip.value_to_string();
            if let Some(ip) = geo.convert(ip.as_str()) {
                return match key {
                    "SrcASN" => geo.get_asn(ip),
                    "SrcCity" => geo.get_city(ip),
                    _ => geo.get_country(ip),
                };
            }
        }
        // No GeoIP DB, missing sourceIPAddress, or a non-IP value
        // (e.g. "cloudtrail.amazonaws.com"): emit a placeholder, never the raw string.
        return "-".to_string();
    }
```

This is behavior-preserving for every case except the bug: for a valid IP the three geo columns still return their looked-up values; non-geo columns are no longer touched by the GeoIP block and continue to the normal field-processing path below; and geo columns with no DB / missing / unparseable IP now consistently return `-` (matching the placeholder already used elsewhere in the function). The change is local to `get_value_from_event_common` — no shared helper is needed.

## Verification

A behavioral unit test of `get_value_from_event_common` requires a live `GeoIPSearch`, which needs the MaxMind `.mmdb` files to construct (`GeoIPSearch::new`), so it is not directly unit-testable without a DB fixture. Two complementary checks:

1. **Manual check (primary):** create a one-record CloudTrail input whose `sourceIPAddress` is `cloudtrail.amazonaws.com` and run `suzaku aws-ct-timeline <dir> -G <maxmind-db-dir>`. Before the fix, every column of the output row equals `cloudtrail.amazonaws.com`; after the fix, `Timestamp`/`EventName`/`RuleTitle`/etc. hold their correct values and `SrcASN`/`SrcCity`/`SrcCountry` read `-`.

2. **Unit test (make the parse precondition testable):** if `convert` is lifted to a free/associated function that does not borrow `self` (it currently only calls `IpAddr::from_str` and ignores the readers), a `#[cfg(test)]` test in `src/option/geoip.rs` can lock in that non-IP service principals are rejected — the precondition the bug depended on:

```rust
    #[test]
    fn convert_rejects_non_ip_source_values() {
        // CloudTrail sets sourceIPAddress to service principals for AWS-initiated
        // events; these must not be treated as addresses.
        for s in [
            "cloudtrail.amazonaws.com",
            "AWS Internal",
            "ecs-tasks.amazonaws.com",
            "health.amazonaws.com",
        ] {
            assert!(GeoIPSearch::convert(s).is_none(), "{s} must not parse as an IP");
        }
        assert!(GeoIPSearch::convert("203.0.113.7").is_some());
    }
```

---

Shirofune-Security:
## Summary

Closes #145.

With `--geo-ip`, the GeoIP block in `get_value_from_event_common` ran for **every** output column and, when `sourceIPAddress` wasn't a parseable IP (routine for AWS-service events like `cloudtrail.amazonaws.com`, `AWS Internal`, etc.), its `else { return ip; }` branch returned the raw string **regardless of which column was being computed** — so `Timestamp`, `EventName`, `RuleTitle`, `Level`, and every other column of that row got clobbered with the service-principal string. Exactly the AWS-service events an analyst most wants to read came out corrupted.

## Fix

Scope the whole GeoIP block to only the three enrichment columns:

```rust
if matches!(key, "SrcASN" | "SrcCity" | "SrcCountry") {
    if let Some(geo) = geo_ip && let Some(ip) = event.get("sourceIPAddress") {
        let ip = ip.value_to_string();
        if let Some(ip) = geo.convert(ip.as_str()) {
            return match key { "SrcASN" => geo.get_asn(ip), "SrcCity" => geo.get_city(ip), _ => geo.get_country(ip) };
        }
    }
    return "-".to_string();
}
```

Now a missing DB / missing `sourceIPAddress` / non-IP value yields `-` **only** in the three geo columns; every other column falls through to normal field processing and is never overwritten. The valid-IP path is unchanged.

## Testing

`cargo build`, `clippy --all-targets -D warnings`, `fmt --check`, and `cargo test` all pass. CHANGELOG (EN + JA) updated. (The full `--geo-ip` path needs a MaxMind `.mmdb` to exercise end-to-end, so this is verified by construction — non-geo keys no longer enter the GeoIP block at all — plus the existing suite.)


Shirofune-Security:
Added the regression test you suggested. `geoip_non_ip_source_only_affects_geo_columns` runs with `--geo-ip` actually enabled (`geo_ip = Some(GeoIPSearch)`) against `sourceIPAddress = "cloudtrail.amazonaws.com"` and asserts:

- a normal column (`.eventName`) keeps its own value (`ListBuckets`) — i.e. it is **not** overwritten by the non-IP source, which is exactly the reintroduction it guards against; and
- the GeoIP columns (`SrcCountry`, `SrcASN`) return the `-` placeholder.

Since the test needs a real `GeoIPSearch`, it required GeoLite2 databases — I added the same small (64K total) GeoLite2 test `.mmdb` fixtures Hayabusa ships, under `test_files/mmdb/`, so the test runs in CI rather than skipping. `clippy --all-targets -D warnings`, `fmt`, and `cargo test` all pass locally.

YamatoSecurity (approved):

PR #157, opened 2026-07-18
Validate time filters up front; fix parse_offset panic (#150)

### LINKED ISSUE #150: Malformed --timeline-start/--timeline-end/--time-offset silently drops ALL events (or panics)
The time-based event filters (`--timeline-start`, `--timeline-end`, `--time-offset`) are parsed lazily, once per event, inside `filter_by_time` in `src/option/timefiler.rs`. If a bound cannot be parsed, the function returns `false` for that event — and because the same unparseable bound fails for *every* event, the entire scan produces zero results with no error and no warning. A user who passes a plain date like `--timeline-start 2024-01-01` (instead of full RFC3339 `2024-01-01T00:00:00Z`) gets an empty timeline and no indication anything went wrong. Separately, `parse_offset` computes a byte split index from the *untrimmed* length and applies it to the *trimmed* string, which panics on an empty offset, a multibyte trailing character, or trailing whitespace. In a forensic tool, empty output reads as "nothing happened," so a silent typo-induced drop is a genuine correctness / data-loss hazard.

## Problem

Unlike `--file-date-from` / `--file-date-to`, which are validated up front by a clap `value_parser` (`parse_file_date`), the three time options are declared as bare `Option<String>` with **no** `value_parser`, so malformed values are accepted at the CLI layer and only "handled" later, per event.

`src/option/cli.rs:53-66`:

```rust
#[derive(Args, Clone, Debug, Default)]
pub struct TimeOption {
    /// Start time of the events to load (ex: "2022-02-22T23:59:59Z)
    #[arg(help_heading = Some("Filtering"), long = "timeline-start", value_name = "DATE", display_order = 210)]
    pub timeline_start: Option<String>,

    /// End time of the events to load (ex: "2020-02-22T00:00:00Z")
    #[arg(help_heading = Some("Filtering"), long = "timeline-end", value_name = "DATE", display_order = 211)]
    pub timeline_end: Option<String>,

    /// Scan recent events based on an offset (ex: 1y, 3M, 30d, 24h, 30m)
    #[arg(help_heading = Some("Filtering"), long = "time-offset", value_name = "OFFSET", conflicts_with = "timeline_start", display_order = 212)]
    pub time_offset: Option<String>,
}
```

The bound is then re-parsed for every event in `filter_by_time`, and a parse error silently drops the event.

`src/option/timefiler.rs:37-64`:

```rust
    if let Some(start) = &opt.timeline_start {
        let start_time = match DateTime::parse_from_rfc3339(start) {
            Ok(dt) => dt.with_timezone(&Utc),
            Err(_) => return false,
        };
        if event_time < start_time {
            return false;
        }
    }
    if let Some(end) = &opt.timeline_end {
        let end_time = match DateTime::parse_from_rfc3339(end) {
            Ok(dt) => dt.with_timezone(&Utc),
            Err(_) => return false,
        };
        if event_time > end_time {
            return false;
        }
    }
    if let Some(offset) = &opt.time_offset {
        let now = Utc::now();
        let duration = match parse_offset(offset) {
            Some(d) => d,
            None => return false,
        };
        if event_time < now - duration {
            return false;
        }
    }
```

The `Err(_) => return false` arms mean an invalid start or end bound rejects every event. The `None => return false` arm does the same for an unrecognized offset unit (e.g. `--time-offset 5w`).

`parse_offset` has a second, worse failure mode — it panics.

`src/option/timefiler.rs:115-126`:

```rust
fn parse_offset(offset: &str) -> Option<Duration> {
    let (num, unit) = offset.trim().split_at(offset.len() - 1);
    let n: i64 = num.parse().ok()?;
    match unit {
        "y" => Some(Duration::days(n * 365)),
        "M" => Some(Duration::days(n * 30)),
        "d" => Some(Duration::days(n)),
        "h" => Some(Duration::hours(n)),
        "m" => Some(Duration::minutes(n)),
        _ => None,
    }
}
```

`split_at` is called on `offset.trim()` but the index is `offset.len() - 1` — the length of the **untrimmed** string. This panics when:
- **empty / whitespace-only** offset: `offset.len() - 1` underflows `usize` (`0 - 1`), or the index exceeds the trimmed length → `split_at` panics with out-of-bounds.
- **trailing whitespace** (e.g. `"30d "`): untrimmed length is larger than the trimmed string, so the index lands past the end of the trimmed slice → out-of-bounds panic.
- **multibyte trailing char** (e.g. `"30日"`): the byte index is not a char boundary → `split_at` panics.

## Impact

- **Scope:** any local run of suzaku's timeline commands with a mistyped time bound. This is a locally invoked forensic CLI, so this is an operator-experience / correctness bug, not a remote-exploit path.
- **Silent data loss (primary concern):** an easy, natural typo — `--timeline-start 2024-01-01`, `--timeline-end "2024-01-01 00:00"`, or `--time-offset 7w` — yields an **empty** timeline with exit success and no diagnostic. In an investigation, an empty CSV/JSON output is routinely read as "no matching activity in the window," which is exactly the wrong conclusion to draw from a silently-broken filter. The parsed data is attacker-influenceable cloud-log content, but the trigger here is operator input, not the log.
- **Crash:** `--time-offset ""`, `--time-offset "30d "` (trailing space), or a non-ASCII unit panics the process. It aborts the scan rather than producing a clear error.

Concrete repro:

```
# Silent empty output — no error, exit 0:
suzaku aws-detect -d ./logs --timeline-start 2024-01-01

# Panic:
suzaku aws-detect -d ./logs --time-offset "30d "
```

## Suggested fix

Two minimal, complementary changes: (1) validate the time inputs **once, up front** with clap `value_parser`s so a bad value fails fast with a clear message (mirroring `parse_file_date`), and (2) make `parse_offset` split on the trimmed string safely so it can never panic.

**`src/option/cli.rs`** — add validators and wire them onto the three args:

```rust
/// Validate that the input is a full RFC3339 datetime (e.g. 2024-01-01T00:00:00Z).
fn parse_rfc3339(s: &str) -> Result<String, String> {
    chrono::DateTime::parse_from_rfc3339(s).map_err(|_| {
        format!("'{s}' is not a valid RFC3339 datetime (e.g. 2024-01-01T00:00:00Z)")
    })?;
    Ok(s.to_string())
}

/// Validate a time offset like 1y, 3M, 30d, 24h, 30m.
fn parse_time_offset(s: &str) -> Result<String, String> {
    let t = s.trim();
    let unit = t
        .chars()
        .last()
        .ok_or_else(|| "time offset is empty".to_string())?;
    let num = &t[..t.len() - unit.len_utf8()];
    num.parse::<i64>()
        .map_err(|_| format!("'{num}' is not a valid number in the time offset '{s}'"))?;
    if !matches!(unit, 'y' | 'M' | 'd' | 'h' | 'm') {
        return Err(format!(
            "'{unit}' is not a valid time offset unit (use y, M, d, h, or m)"
        ));
    }
    Ok(s.to_string())
}
```

```diff
     #[arg(help_heading = Some("Filtering"), long = "timeline-start", value_name = "DATE", display_order = 210)]
+    #[arg(value_parser = parse_rfc3339)]
     pub timeline_start: Option<String>,

     #[arg(help_heading = Some("Filtering"), long = "timeline-end", value_name = "DATE", display_order = 211)]
+    #[arg(value_parser = parse_rfc3339)]
     pub timeline_end: Option<String>,

     #[arg(help_heading = Some("Filtering"), long = "time-offset", value_name = "OFFSET", conflicts_with = "timeline_start", display_order = 212)]
+    #[arg(value_parser = parse_time_offset)]
     pub time_offset: Option<String>,
```

**`src/option/timefiler.rs:115-126`** — rewrite `parse_offset` to be panic-free (rebind to the trimmed string before indexing, split off the last `char` by its UTF-8 length):

```rust
fn parse_offset(offset: &str) -> Option<Duration> {
    let offset = offset.trim();
    let unit = offset.chars().last()?;
    let num = &offset[..offset.len() - unit.len_utf8()];
    let n: i64 = num.parse().ok()?;
    match unit {
        'y' => Some(Duration::days(n * 365)),
        'M' => Some(Duration::days(n * 30)),
        'd' => Some(Duration::days(n)),
        'h' => Some(Duration::hours(n)),
        'm' => Some(Duration::minutes(n)),
        _ => None,
    }
}
```

With `parse_offset` no longer able to panic and the clap validators rejecting bad input at startup, the per-event `Err(_) => return false` / `None => return false` arms in `filter_by_time` become unreachable defensive fallbacks — they can stay as-is.

## Verification

Add unit tests in the existing `#[cfg(test)] mod tests` block in `src/option/timefiler.rs`. `parse_offset` is module-private, so it is reachable via `super::*` (already imported):

```rust
#[test]
fn test_parse_offset_valid_units() {
    assert_eq!(parse_offset("30d"), Some(Duration::days(30)));
    assert_eq!(parse_offset("24h"), Some(Duration::hours(24)));
    assert_eq!(parse_offset("1y"), Some(Duration::days(365)));
}

#[test]
fn test_parse_offset_trailing_whitespace_does_not_panic() {
    // Previously panicked: untrimmed len used to index the trimmed string.
    assert_eq!(parse_offset("30d "), Some(Duration::days(30)));
}

#[test]
fn test_parse_offset_empty_does_not_panic() {
    // Previously panicked on `offset.len() - 1` underflow.
    assert_eq!(parse_offset(""), None);
    assert_eq!(parse_offset("   "), None);
}

#[test]
fn test_parse_offset_multibyte_last_char_does_not_panic() {
    // Previously panicked on a non-char-boundary split.
    assert_eq!(parse_offset("30日"), None);
}

#[test]
fn test_parse_offset_unknown_unit() {
    assert_eq!(parse_offset("5w"), None);
}
```

Manual check for the fail-fast behavior: `suzaku aws-detect -d ./logs --timeline-start 2024-01-01` should now exit non-zero with a clear "not a valid RFC3339 datetime" message instead of silently emitting an empty timeline, and `--time-offset "30d "` should no longer panic.

---

Shirofune-Security:
## Summary

Closes #150.

The time filters (`--timeline-start`, `--timeline-end`, `--time-offset`) were bare `Option<String>` with **no** `value_parser`, so a malformed value was accepted at the CLI and only "handled" later, re-parsed once per event in `filter_by_time` — where a parse error returns `false`. Since the same bad bound fails for *every* event, the whole scan produced **zero results with no error or warning**. A user passing a plain `--timeline-start 2024-01-01` (instead of RFC 3339 `2024-01-01T00:00:00Z`) got an empty timeline and no indication anything was wrong — a genuine data-loss hazard in a forensic tool.

Separately, `parse_offset` computed a byte split index from the **untrimmed** length and applied it to the **trimmed** string, panicking on an empty offset, trailing whitespace, or a multibyte trailing character.

## Fix

- Add `value_parser`s (`parse_time_bound` for RFC 3339, `parse_time_offset` reusing `parse_offset`) to the three options, matching the existing `--file-date-from/-to` pattern. Bad values are now rejected at parse time with a clear message.
- Rewrite `parse_offset` to split the trailing unit via `chars().next_back()` — no panic on empty/whitespace/multibyte input.

## Verification

- New unit tests for `parse_offset` (valid units incl. trimmed; empty / whitespace-only / multibyte / bad-unit all return `None` without panicking).
- End-to-end:
  - `--timeline-start 2024-01-01` → `error: invalid value '2024-01-01' for '--timeline-start <DATE>': ... not a valid RFC 3339 date-time ...` (exit 2), instead of an empty timeline.
  - `--time-offset ''` → clean error instead of a panic.
  - Valid `--timeline-start 2020-01-01T00:00:00Z` and `--time-offset 3M` still work.
- `cargo build`, `clippy --all-targets -D warnings`, `fmt --check`, `cargo test` all pass. CHANGELOG EN + JA updated.


Shirofune-Security:
Good catch — fixed in the latest commit. `parse_offset` now uses `checked_mul` for the `y`/`M` conversions and chrono's non-panicking `try_days`/`try_hours`/`try_minutes` (the plain `Duration::days`/`hours`/`minutes` also panic on out-of-range values, so `9223372036854775807d`/`h`/`m` were affected too, not just the multiplied units). An overflowing offset now returns `None`, so the clap value parser rejects it cleanly:

```
$ suzaku aws-ct-timeline -f test_files/json/DeleteTrail.json --time-offset 9223372036854775807y
error: invalid value '9223372036854775807y' for '--time-offset <OFFSET>': '9223372036854775807y' is not a valid time offset (e.g. 1y, 3M, 30d, 24h, 30m)
```

Added a regression test covering `i64::MAX`/`i64::MIN` for `y`/`M`/`d`/`h`/`m`. `clippy --all-targets -D warnings`, `fmt`, and `cargo test` all pass.

YamatoSecurity (approved):

Issue #147, opened 2026-07-17
Unbounded gzip decompression allows a decompression-bomb OOM (aborts the scan)


Shirofune-Security:
`read_gz_file` in `src/core/scan.rs` decompresses `.gz` inputs fully into memory with no cap on the decompressed size. Because gzip can inflate at roughly 1032:1, a small crafted `.gz` file (a few MB) can expand to many gigabytes and exhaust memory, causing the process to be OOM-killed. Since `.gz` files are picked up automatically during a directory scan and the decompression happens on the main scan path (as well as in the AWS metrics/summary/search commands), a single malicious or corrupt archive dropped into a log directory aborts the entire run.

## Problem

`read_gz_file` reads the whole decompressed stream into a `String` with no size limit:

```rust
// src/core/scan.rs:605-611
pub fn read_gz_file(file_path: &PathBuf) -> io::Result<String> {
    let file = File::open(file_path)?;
    let mut decoder = GzDecoder::new(BufReader::new(file));
    let mut contents = String::new();
    decoder.read_to_string(&mut contents)?;
    Ok(contents)
}
```

`read_to_string` drives the decoder to EOF and grows `contents` to hold the entire decompressed payload. The on-disk size gives no useful bound: the DEFLATE format's maximum expansion ratio is ~1032x, so a ~5 MB archive of repeated bytes decompresses to ~5 GB.

This path is reachable from an ordinary scan:

- `count_files_recursive` whitelists `.gz` alongside `.json`/`.jsonl`/`.csv` when walking the input directory (`src/core/scan.rs:584-585`).
- `get_content` routes any `.gz` path to `read_gz_file` (`src/core/scan.rs:663-664`), and `get_content` is called on the main scan path at `src/core/scan.rs:38` and directly at `src/core/scan.rs:177` (`process_events_from_dir`), plus `aws_metrics.rs:55`, `aws_summary.rs:401`, and `aws_search.rs:120`.

Because the allocation is unbounded, the OS OOM-kills the process before any per-file error handling (`Err(_) => continue` / `unwrap_or_default()`) can skip the bad file, so the whole scan dies rather than just that one input being dropped.

Note: the plain-text branch of `get_content` uses `fs::read_to_string` (`src/core/scan.rs:662`), which is also unbounded, but a `.json`/`.jsonl`/`.csv` file large enough to OOM must actually be that large on disk — there is no decompression amplification, so it is far less practical to weaponize and not the focus here.

## Impact

[...]

PR #158, opened 2026-07-20
Add Parquet input support to AWS commands

nishikawaakira:
## Overview

Adds Parquet (`.parquet`) input support to the AWS commands (`aws-ct-timeline`, `aws-ct-metrics`, `aws-ct-summary`, `aws-ct-search`) for both `-f` and `-d`.

## Implementation

- New `load_parquet_events()` in `src/core/scan.rs` converts each Parquet row into one JSON event via the `parquet` + `arrow-json` crates (v59.1.0). Nested struct columns become nested JSON objects.
- **Athena / Glue / Firehose pipeline shape**: known CloudTrail envelope fields (`userIdentity`, `requestParameters`, `responseElements`, etc.) stored as serialized JSON strings are parsed back into objects so rules can match nested values. Free-text fields such as `errorMessage` are intentionally left alone.
- **Timestamps**: timezone-less `eventTime` TIMESTAMP columns (rendered as `2026-05-03T16:05:07`) are marked as UTC by appending `Z`, since CloudTrail timestamps are always UTC — this keeps `--timeline-start/end` filtering and the date summaries working.
- Compression codecs enabled: snappy, gzip, zstd, lz4.
- The three single-file code paths in `aws-ct-metrics` / `aws-ct-summary` / `aws-ct-search` now share a `load_aws_events_from_file()` helper.

## Testing

- 2 new unit tests with committed fixtures under `test_files/parquet/`: an Athena-style file (envelope fields as JSON strings + naive TIMESTAMP `eventTime`, 29 records) and a nested-struct file. All 122 tests, clippy, and fmt pass.
- End-to-end parity check: a real 2,999-event CloudTrail incident log scanned as JSONL vs. as Parquet produced **identical detections** (1,422 events with hits; zero per-event-per-rule differences), for both `-f` and `-d`, and the other three commands were smoke-tested with Parquet input.

## Out of scope

Athena CTAS lowercases column names (`eventname`, `useridentity`, …); supporting that shape needs a canonical CloudTrail field-name normalization map and is left for a follow-up.


Shirofune-Security:
Reviewed the whole diff — this is clean, well-scoped, and looks correct. A few things I specifically checked and liked:

- The envelope-field revival is limited to the 8 known object/array fields and guarded by `matches!(parsed, Object | Array)`, so free-text like `errorMessage` is left alone even if it contains JSON-looking text. 👍
- The `eventTime`→UTC marking parses as `NaiveDateTime` **first**, so it correctly won't double-append `Z` to a value that already has a `Z`/offset (a `parse_from_str(..., "%Y-%m-%dT%H:%M:%S%.f")` on `...07Z` leaves the trailing `Z` unconsumed → `is_err()` → skipped). Nice.
- In the `-d` loop the `.parquet` branch `continue`s before the `read_to_string` block, so binary Parquet never hits the text path, and it uses the real `PathBuf` (so non-UTF-8 names still resolve). Consistent with the rest of the walk.
- Tests + committed fixtures + the end-to-end JSONL-vs-Parquet parity check are convincing.

**One follow-up worth considering (non-blocking):** `load_parquet_events` decodes the whole file into memory unbounded, and Parquet is a *compressed* format (this PR enables snappy/zstd/gzip/lz4). That's the same decompression-bomb class we just capped for `.gz` in #147 — a small crafted `.parquet` can inflate to many GB and OOM the scan, and `.parquet` files are auto-discovered during a `-d` walk over an attacker-influenceable log tree. It's parity with the (also-unbounded) JSON path, so I wouldn't block on it, but bounding the total decoded rows/bytes (and skipping + warning past the cap, like `read_gz_file_capped`) would close the gap.

Minor, no action needed: for `aws-ct-metrics`/`-summary`/`-search`, `-f` routes through `load_aws_events_from_file`, which handles json/gz/parquet but not `.csv` — that matches the pre-existing behavior of those `-f` paths (they never parsed CSV), so no regression; just noting since `scan_file` (timeline `-f`) does handle `.csv`.

Otherwise LGTM.

fukusuket (approved):
LGTM!!🚀

YamatoSecurity (approved):
@nishikawaakira LGTM! Thank you!

PR #167, opened 2026-07-20
fix: guard summary Data-reduction against underflow/NaN (#163)

### LINKED ISSUE #163: Summary "Data reduction" underflows / prints NaN when correlation double-counts hits
The results-summary header computes "Data reduction" as `total_events - event_with_hits` using unsigned `usize` subtraction and then divides by `total_events` with no zero guard (`src/core/summary.rs:64-65`). However `event_with_hits` is incremented in two independent places — once per hit-bearing event in the base scan (`src/core/scan.rs:465`) and again once per matched correlation event in correlation processing (`src/core/timeline.rs:212`) — while `total_events` is only ever incremented in the base scan (`src/core/scan.rs:469`). Because correlation matches add to `event_with_hits` but not to `total_events`, the invariant `event_with_hits <= total_events` can be violated, causing the unsigned subtraction to underflow: a debug/test panic (`attempt to subtract with overflow`) or, in release builds, a wrapped `18446744073709551615`. On empty input the `0 / 0` division additionally prints `NaN%`.

## Problem

`print_summary_header` in `src/core/summary.rs` (lines 60-68) builds the "Data reduction" string with a raw `usize` subtraction and an unguarded floating-point division:

```rust
p(
    Green.rdg(no_color),
    &format!(
        "Data reduction: {} events ({:.2}%)",
        (sum.total_events - sum.event_with_hits).to_formatted_string(&Locale::en),
        (sum.total_events - sum.event_with_hits) as f64 * 100.0 / sum.total_events as f64
    ),
    false,
);
```

The two summary counters are maintained inconsistently. In the base scan, `src/core/scan.rs:465-469` increments both counters, but from disjoint denominators:

```rust
summary.event_with_hits += results
    .iter()
    .filter(|(_, _, matched_rules)| !matched_rules.is_empty())
    .count();
summary.total_events += json_events.len();
```

Then correlation processing bumps `event_with_hits` a second time, per matched correlation event, without ever touching `total_events` (`src/core/timeline.rs:204-213`):

```rust
for res in results.iter() {
    let rule = res.rule;
    if res.matched {
        for event in &res.events {
            let generate = rule.correlation.generate.unwrap_or(false);
            if generate {
                write_record(&event.event, &Value::Null, Some(event.rule), context);
            }
            summary.event_with_hits += 1;
            append_summary_data(summary, &event.event, event.rule, generate, context);
        }
```

Since `total_events` counts distinct input events but `event_with_hits` now counts base-rule hits *plus* correlation-rule hits (the same underlying event can be counted by a base rule and one or more correlation rules), `event_with_hits` can exceed `total_events`. When it does, `sum.total_events - sum.event_with_hits` underflows. In debug/test builds Rust's overflow checks panic; in release builds the subtraction wraps around to a value near `1.8e19`, and the percentage becomes a nonsensical large or negative-then-wrapped figure. Separately, with no input at all both counters are `0`, so the division `0.0 * 100.0 / 0.0` yields `NaN` and the header prints `Data reduction: 0 events (NaN%)`.

## Impact

Suzaku is a locally-run forensic timeline CLI, so this is not remotely exploitable, but the counters are driven by attacker-influenceable cloud-log content: whether a given CloudTrail/Azure/M365 event matches a base rule and one or more correlation rules is a function of the log values an adversary can shape. The concrete failure modes are:

- **Debug/test builds panic.** Any `cargo test` or debug-profile run over a dataset where a single event is matched by both a base rule and a matching correlation rule aborts the summary with `attempt to subtract with overflow`, which can mask other output and break CI.
- **Release builds print garbage.** The same dataset produces `Data reduction: 18446744073709551615 events (...%)` in the terminal summary, undermining trust in the tool's headline statistic during an investigation.
- **Empty/all-filtered input prints `NaN%`.** Running against an empty log set or a time window that excludes everything yields `Data reduction: 0 events (NaN%)`.

Concrete repro: feed one event that is matched by a base detection rule and also referenced by exactly one matching correlation rule. Base scan sets `total_events = 1`, `event_with_hits = 1`; correlation processing then does `event_with_hits += 1`, giving `event_with_hits = 2 > total_events = 1`. `print_summary_header` then evaluates `1 - 2` → panic (debug) or `18446744073709551615` (release).

## Suggested fix

Make the header arithmetic total-order-safe with `saturating_sub` and guard the percentage against a zero denominator. This is the minimal change that removes the panic/underflow/NaN regardless of counter accounting:

```rust
let reduced = sum.total_events.saturating_sub(sum.event_with_hits);
let reduction_pct = if sum.total_events == 0 {
    0.0
} else {
    reduced as f64 * 100.0 / sum.total_events as f64
};
p(
    Green.rdg(no_color),
    &format!(
        "Data reduction: {} events ({:.2}%)",
        reduced.to_formatted_string(&Locale::en),
        reduction_pct
    ),
    false,
);
```

Longer term, the root cause is the double-counting: correlation matches inflate `event_with_hits` without a corresponding `total_events` denominator, so even with `saturating_sub` the reported "Data reduction" can read `0 events (0.00%)` when correlation hits outnumber base hits. The cleaner fix is to stop counting correlation events into `event_with_hits` (they are derived from already-counted base events), i.e. drop the `summary.event_with_hits += 1;` at `src/core/timeline.rs:212` and track correlation counts in a separate field if they need to be surfaced — preserving the invariant `event_with_hits <= total_events`. The `saturating_sub` + zero-guard above should land regardless, as defense in depth.

## Verification

Add a unit test in `src/core/summary.rs` under a `#[cfg(test)]` module that constructs a `DetectionSummary` with the pathological counts and asserts the header renders without panicking and without underflow/NaN. Because `print_summary_header` currently prints rather than returns, either assert on a refactored helper that returns the tuple, or (minimal) assert the arithmetic directly:

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn data_reduction_does_not_underflow_when_hits_exceed_total() {
        // one event matched by a base rule AND one correlation rule
        let total_events: usize = 1;
        let event_with_hits: usize = 2;

        let reduced = total_events.saturating_sub(event_with_hits);
        let pct = if total_events == 0 {
            0.0
        } else {
            reduced as f64 * 100.0 / total_events as f64
        };

        assert_eq!(reduced, 0);
        assert!(pct.is_finite());
    }

    #[test]
    fn data_reduction_empty_input_is_not_nan() {
        let total_events: usize = 0;
        let event_with_hits: usize = 0;

        let reduced = total_events.saturating_sub(event_with_hits);
        let pct = if total_events == 0 {
            0.0
        } else {
            reduced as f64 * 100.0 / total_events as f64
        };

        assert_eq!(reduced, 0);
        assert!(pct.is_finite(), "percentage must not be NaN on empty input");
    }
}
```

The first test reproduces the panic against the *current* code (unsigned `1 - 2`) and passes once `saturating_sub` is in place; the second pins the `NaN%` regression. Manual check: run a release build over a dataset containing one event matched by both a base rule and a matching correlation rule and confirm the summary no longer prints `18446744073709551615` or `NaN%`.

---

Shirofune-Security:
Fixes #163.

## Problem
`print_summary_header` (`src/core/summary.rs`) computed the "Data reduction" line as `(total_events - event_with_hits)` with unsigned `usize` subtraction and divided by `total_events` with no zero guard. `event_with_hits` is incremented in the base scan (per event with ≥1 hit) **and again** in correlation processing (`timeline.rs`, per event per matched correlation rule) for events already counted, while `total_events` is only incremented in the base scan — so `event_with_hits` can exceed `total_events`.

When it does: **panic** (`attempt to subtract with overflow`) in debug/tests, a wrapped **~1.8e19** count in release, or **`NaN%`** on empty input.

## Fix
Extracted the arithmetic into a pure `data_reduction(total, with_hits) -> (usize, f64)` helper using `saturating_sub` and an explicit zero guard, and used it in the header. Unit-tested the double-count and empty cases.

## Testing
`cargo test data_reduction` passes; full `cargo test` unaffected.


fukusuket (approved):
LGTM!!🚀

PR #168, opened 2026-07-20
fix: track per-entry first/last-seen in aws-ct-summary (#160)

### LINKED ISSUE #160: aws-ct-summary: per-entry first_seen/last_seen hold the global timestamp, not the entry's own range
In `aws-ct-summary`, the per-key summary maps (regions, source IPs, APIs, access-key IDs, user-agents) store a `first_seen`/`last_seen` for each entry, but these fields are frozen at the moment the key is first inserted and are seeded from the **dataset-global** `first_timestamp`/`last_timestamp` — not from the entry's own occurrences. On every subsequent hit for the same key only the count is incremented; the timestamps are never revisited. As a result the CSV and JSON output report the global min/max for every row instead of that entry's actual first and last occurrence, making the per-entry timeline columns wrong and misleading during a forensic investigation.

## Problem

In `src/cmd/aws/aws_summary.rs`, `CTSummary::add_event` (lines 80-158) first updates the global timestamps and then populates each per-key map with `.or_insert(...)` seeded from those global values, incrementing only the count `.0` on every hit:

```rust
if self.first_timestamp.is_empty() || event_time < self.first_timestamp {
    self.first_timestamp = event_time.clone();
}
if self.last_timestamp.is_empty() || event_time > self.last_timestamp {
    self.last_timestamp = event_time.clone();
}

let entry = self.aws_regions.entry(aws_region.clone()).or_insert((
    0,
    self.first_timestamp.clone(),
    self.last_timestamp.clone(),
));
entry.0 += 1;
let entry = self.src_ips.entry(source_ip.clone()).or_insert((
    0,
    self.first_timestamp.clone(),
    self.last_timestamp.clone(),
));
entry.0 += 1;
self.user_types = user_type.clone();
let entry = self.access_key_ids.entry(access_key_id.clone()).or_insert((
    0,
    self.first_timestamp.clone(),
    self.last_timestamp.clone(),
));
entry.0 += 1;
let entry = self.user_agents.entry(user_agent.clone()).or_insert((
    0,
    self.first_timestamp.clone(),
    self.last_timestamp.clone(),
));
entry.0 += 1;

if !abused_api_success.is_empty() {
    let entry = self
        .abused_api_success
        .entry(abused_api_success.clone())
        .or_insert((0, self.first_timestamp.clone(), self.last_timestamp.clone()));
    entry.0 += 1;
}
// ...abused_api_failed / other_api_success / other_api_failed follow the same pattern
```

Two mechanisms combine into the bug:

1. **Wrong seed value.** The tuple's timestamp fields (`.1`/`.2`) are initialized from `self.first_timestamp`/`self.last_timestamp` — the running global min/max — rather than from this event's own `event_time`.
2. **Never updated on re-hit.** `.or_insert(...)` only runs its argument when the key is absent, and the code that follows touches only `.0`. So once a key exists, its `.1`/`.2` are permanently whatever the global window happened to be when the key was first seen; later occurrences of the same key never widen the entry's own range.

These tuples flow straight into the output via `map_to_api_entries` and `map_to_count_entries` (lines 167-201), which read `first`/`last` from `.1`/`.2`, so both the JSON `first_seen`/`last_seen` and the corresponding CSV columns are affected.

## Impact

`suzaku` is a locally-run forensic timeline/summary CLI, so this is not a memory-safety or injection issue — it is output correctness. The affected columns are exactly the per-entity timeline data an analyst relies on: "when did this source IP / access key / API call / user-agent first and last appear for this ARN." Because CloudTrail field values (source IP, user-agent, region, etc.) are attacker-influenceable, and the timestamps themselves are attacker-controlled to the extent an attacker paces their activity, an investigator reading the summary can be led to a wrong conclusion about when a specific indicator was active — e.g. an IP that only appeared once at the very start of the window will be shown with a `last_seen` equal to the global end of the dataset, and vice-versa. Analysts open the CSV in spreadsheets and sort/filter on these columns, amplifying the wrong values.

Concrete repro: build a dataset with three events for one user ARN:

- `T1 = 2024-01-01T00:00:00Z` — API `A` from IP `1.1.1.1`
- `T2 = 2024-01-02T00:00:00Z` — API `B` from IP `2.2.2.2`
- `T3 = 2024-01-03T00:00:00Z` — API `A` from IP `1.1.1.1`

Expected for entry `1.1.1.1`: `first_seen = T1`, `last_seen = T3`. Expected for entry `2.2.2.2`: `first_seen = last_seen = T2`. Actual: `1.1.1.1` is inserted at T1 when the global window is `[T1, T1]`, so it is frozen at `first=T1, last=T1` and never updated to T3; `2.2.2.2` is inserted at T2 when the global window is `[T1, T2]`, so it reports `first=T1, last=T2` — neither of which is that IP's real range.

## Suggested fix

Seed each entry from the event's own `event_time`, and widen `.1`/`.2` with a lexicographic min/max on every hit (string compare is correct here because the values are the same ISO-8601 UTC strings already relied on for the global fields). A small free helper keeps the six call sites consistent and avoids repetition:

```rust
/// Increment the count for `key` and widen its first/last-seen window to include `event_time`.
/// ISO-8601 UTC strings sort lexicographically, so string min/max is a correct time comparison.
fn bump(map: &mut HashMap<String, (usize, String, String)>, key: String, event_time: &str) {
    let entry = map
        .entry(key)
        .or_insert_with(|| (0, event_time.to_string(), event_time.to_string()));
    entry.0 += 1;
    if event_time < entry.1.as_str() {
        entry.1 = event_time.to_string();
    }
    if event_time > entry.2.as_str() {
        entry.2 = event_time.to_string();
    }
}
```

Then replace the per-map blocks in `add_event` (the global-timestamp update at the top stays unchanged):

```rust
        bump(&mut self.aws_regions, aws_region, &event_time);
        bump(&mut self.src_ips, source_ip, &event_time);
        self.user_types = user_type;
        bump(&mut self.access_key_ids, access_key_id, &event_time);
        bump(&mut self.user_agents, user_agent, &event_time);

        if !abused_api_success.is_empty() {
            bump(&mut self.abused_api_success, abused_api_success, &event_time);
        }
        if !abused_api_failed.is_empty() {
            bump(&mut self.abused_api_failed, abused_api_failed, &event_time);
        }
        if !other_api_success.is_empty() {
            bump(&mut self.other_api_success, other_api_success, &event_time);
        }
        if !other_api_failed.is_empty() {
            bump(&mut self.other_api_failed, other_api_failed, &event_time);
        }
```

(The `.clone()` calls on the moved arguments can be dropped since each value is now consumed exactly once; `event_time` is borrowed. `bump` can be a free function in this module or an associated `fn` — a free function keeps the borrow of `self`'s individual maps unambiguous.)

## Verification

Add a `#[cfg(test)]` unit test that drives `add_event` with interleaved events for the same key and asserts the per-entry window, which fails on `main` and passes after the fix:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn per_entry_first_last_seen_track_own_occurrences() {
        let mut s = CTSummary::default();
        let ev = |t: &str, ip: &str, api: &str| {
            (t.to_string(), "us-east-1".to_string(), ip.to_string(),
             "IAMUser".to_string(), "AKIA".to_string(), "ua".to_string(),
             "".to_string(), "".to_string(), api.to_string(), "".to_string())
        };
        for (t, ip, api) in [
            ev("2024-01-01T00:00:00Z", "1.1.1.1", "A"),
            ev("2024-01-02T00:00:00Z", "2.2.2.2", "B"),
            ev("2024-01-03T00:00:00Z", "1.1.1.1", "A"),
        ] {
            s.add_event(t.0, t.1, t.2, t.3, t.4, t.5, t.6, t.7, t.8, t.9);
        }

        // Global window spans the whole dataset.
        assert_eq!(s.first_timestamp, "2024-01-01T00:00:00Z");
        assert_eq!(s.last_timestamp, "2024-01-03T00:00:00Z");

        // Entry seen twice: window is its own first..last, not the global window.
        let ip1 = &s.src_ips["1.1.1.1"];
        assert_eq!(ip1.0, 2);
        assert_eq!(ip1.1, "2024-01-01T00:00:00Z");
        assert_eq!(ip1.2, "2024-01-03T00:00:00Z");

        // Entry seen once in the middle: first_seen == last_seen == its own time.
        let ip2 = &s.src_ips["2.2.2.2"];
        assert_eq!(ip2.0, 1);
        assert_eq!(ip2.1, "2024-01-02T00:00:00Z");
        assert_eq!(ip2.2, "2024-01-02T00:00:00Z");
    }
}
```

Run with `cargo test per_entry_first_last_seen_track_own_occurrences`. On `main`, `ip2.1` is `"2024-01-01T00:00:00Z"` (global min) and `ip1.2` is `"2024-01-01T00:00:00Z"` (frozen at insertion), so the assertions fail; after the fix they pass. A manual check works too: run `suzaku aws-ct-summary` over any multi-event dataset and confirm an indicator that appears only at the edge of the window no longer shows the global min/max in its `first_seen`/`last_seen`.

---

Shirofune-Security:
Fixes #160.

## Problem
`CTSummary::add_event` (`src/cmd/aws/aws_summary.rs`) populated each per-key map (regions, source IPs, access keys, user agents, and the abused/other API maps) with `.or_insert((0, self.first_timestamp.clone(), self.last_timestamp.clone()))` and then only did `entry.0 += 1` on later hits. So every entry's `first_seen`/`last_seen`:
1. was seeded from the **dataset-global** running min/max at the moment that key was first seen (not the event's own time), and
2. was **never updated** afterward.

The per-entry time windows in both the CSV and JSON output are therefore wrong — they show (a snapshot of) the whole dataset's span instead of that region/IP/key's actual first and last occurrence.

## Fix
Added `upsert_count_entry(map, key, event_time)` which seeds the tuple from the **event's own** timestamp and updates `min`/`max` (lexicographic on the ISO-8601 strings, matching the existing global-timestamp logic) on every hit, and routed all eight per-key maps through it. The dataset-global `first_timestamp`/`last_timestamp` (used for the summary's overall range) are unchanged.

## Testing
`cargo test upsert_count_entry_tracks_per_key_time_range` — asserts a key seen at T1/T3/T2 ends with `first_seen=T1`, `last_seen=T3`, `count=3`. Full `cargo test` unaffected.


Shirofune-Security:
CI is failing in `cargo fmt -- --check`. The only reported issue is formatting in `src/cmd/aws/aws_summary.rs` around the new `Self::upsert_count_entry(&mut self.abused_api_success, abused_api_success, &event_time);` call. Please run `cargo fmt` on the branch and push the formatted result.

fukusuket (approved):
LGTM!!🚀

PR #169, opened 2026-07-20
fix: harden aws-ct-search --regex/--filter and abused-API warning (#162)

### LINKED ISSUE #162: aws-ct-* robustness: --regex panics on a bad pattern, --filter is silently dropped, and abused-API classification depends on a cwd-relative file
Three related robustness gaps in the AWS CloudTrail subcommands degrade the user experience and, in one case, silently corrupt the `aws-ct-summary` output. `aws-ct-search --regex` compiles the user's pattern with `.expect(...)`, so a bad pattern aborts the whole run with a Rust panic and backtrace instead of a clean fatal error. `--filter` values that are not in `FIELD:VALUE` form are silently discarded by `parse_filter_conditions`, so a typo'd filter is applied as "no filter" with no feedback. And `aws-ct-summary` loads its abused-API reference list from the hard-coded, current-working-directory-relative path `rules/config/abused_aws_api_calls.csv`; when suzaku is launched from any other directory the file is not found, the failure is non-fatal, and every API call is then bucketed as non-abused — quietly skewing the abused/non-abused split of the summary. None of this overlaps the already-merged #145–#150 fixes.

## Problem

**1. `--regex` panics; `--filter` is silently dropped** — `src/cmd/aws/aws_search.rs:48-51`

```rust
let regex_pattern = options
    .regex
    .as_ref()
    .map(|pattern| Regex::new(pattern).expect("Invalid regex pattern"));
```

`Regex::new` returns `Result<Regex, regex::Error>`; `.expect()` turns any invalid user-supplied pattern (e.g. an unclosed group `(`) into a `panic!`, printing a Rust backtrace rather than the clean, actionable message that `fatal_error` (`src/core/util.rs:60`) exists to provide everywhere else in the input path.

Separately, `--filter` values are parsed at `src/cmd/aws/aws_search.rs:46` via `parse_filter_conditions` (`src/cmd/aws/aws_search.rs:139-148`):

```rust
fn parse_filter_conditions(filters: &[String]) -> Vec<(String, String)> {
    let mut conditions = Vec::new();
    for filter in filters {
        if let Some((field, value)) = filter.split_once(':') {
            let value = value.trim_matches('"').trim_matches('\'');
            conditions.push((field.to_string(), value.to_string()));
        }
    }
    conditions
}
```

Any filter string lacking a `:` fails the `if let Some(...)` and is silently skipped. A user who writes `--filter eventName=ConsoleLogin` (or otherwise fat-fingers the separator) gets a run that returns *more* events than intended, with no warning that their filter was ignored.

**2. Abused-API classification depends on a cwd-relative file** — `src/cmd/aws/aws_summary.rs:268`

```rust
let abused_aws_api_calls = read_abused_aws_api_calls("rules/config/abused_aws_api_calls.csv");
```

The literal is relative to the process's current working directory, and `rules/` is not part of the suzaku binary — it is fetched separately (the rules repo). `read_abused_aws_api_calls` (`src/cmd/aws/aws_summary.rs:636-661`) handles a missing file by printing one terse line and returning an empty map:

```rust
Err(_) => {
    p(
        Some(Color::Rgb(255, 0, 0)),
        "Failed to open the abused AWS API calls file.",
        true,
    );
    HashMap::new()
}
```

The message does not name the path it tried, does not explain the consequence, and does not stop the run. With an empty map, the per-event classification at `src/cmd/aws/aws_summary.rs:337-359` never matches: `abused_aws_api_calls.get(&event_name)` is always `None`, `.contains_key(&event_name)` is always `false`, so *every* call falls into the `other_api_success` / `other_api_failed` (non-abused) buckets. The summary's headline abused-vs-non-abused breakdown collapses to "nothing was abused," which is exactly the signal an analyst is looking at the report for.

## Impact

suzaku is a locally-run forensic CLI, so none of this is remotely exploitable — but it directly undermines the tool's job.

- **`--regex` panic**: any typo in the pattern (e.g. `suzaku aws-ct-search -d ./logs --regex '(unclosed'`) ends the run with a `thread 'main' panicked ... Invalid regex pattern` backtrace instead of a one-line error, which reads as a suzaku crash rather than user input error.
- **`--filter` silent drop**: `suzaku aws-ct-search -d ./logs --filter 'eventName=ConsoleLogin'` runs to completion and prints results as if filtering succeeded, but the filter was ignored — a false-negative-free result set that the analyst believes was scoped. Because CloudTrail field values are attacker-influenceable, an over-broad result set can also bury the events of interest.
- **abused-API misclassification (most serious)**: `cd /somewhere-else && suzaku aws-ct-summary -d /path/to/logs` — or any invocation not started from the directory that contains `rules/config/` — produces a summary in which the abused-API section is empty and 100% of activity is labeled non-abused. Since summary output is commonly exported to CSV and opened in a spreadsheet for triage, a reviewer scanning the "abused APIs" column sees nothing and may conclude no known-abused API calls occurred, when in fact classification never ran. The single red line scrolls past above pages of normal output and is easy to miss.

Concrete repro for the third issue:
```
$ cd /tmp && suzaku aws-ct-summary -d /path/to/cloudtrail-logs -o out.csv
Failed to open the abused AWS API calls file.   # easily missed
$ # out.csv: abused APIs sections empty, every event counted as "other"
```

## Suggested fix

Compile `--regex` and validate `--filter` through the existing `fatal_error` path, and resolve the abused-API CSV relative to the executable (falling back to cwd) with a loud, consequence-stating warning when it is genuinely absent.

**`src/cmd/aws/aws_search.rs`** — replace the `.expect()` with a clean fatal error:

```rust
let regex_pattern = match options.regex.as_ref() {
    Some(pattern) => match Regex::new(pattern) {
        Ok(re) => Some(re),
        Err(e) => fatal_error(no_color, &format!("Invalid --regex pattern: {e}")),
    },
    None => None,
};
```

Make a malformed `--filter` an explicit error instead of a silent skip (call site becomes `parse_filter_conditions(&options.filter, no_color)`):

```rust
fn parse_filter_conditions(filters: &[String], no_color: bool) -> Vec<(String, String)> {
    let mut conditions = Vec::new();
    for filter in filters {
        match filter.split_once(':') {
            Some((field, value)) => {
                let value = value.trim_matches('"').trim_matches('\'');
                conditions.push((field.to_string(), value.to_string()));
            }
            None => fatal_error(
                no_color,
                &format!("Invalid --filter value '{filter}'. Expected FIELD:VALUE format."),
            ),
        }
    }
    conditions
}
```

**`src/cmd/aws/aws_summary.rs`** — resolve the CSV relative to the executable, and state the consequence on failure. Add a small resolver and switch `read_abused_aws_api_calls` to take a `&Path` (`std::path::Path` is already imported):

```rust
/// Resolve a rules/config asset: prefer the path relative to the current
/// working directory (existing behavior), then fall back to the directory
/// containing the suzaku executable so the tool works when launched from
/// anywhere.
fn resolve_rules_config_path(relative: &str) -> PathBuf {
    let cwd_path = PathBuf::from(relative);
    if cwd_path.exists() {
        return cwd_path;
    }
    if let Ok(exe) = std::env::current_exe()
        && let Some(dir) = exe.parent()
    {
        let exe_path = dir.join(relative);
        if exe_path.exists() {
            return exe_path;
        }
    }
    cwd_path
}
```

Call site (`src/cmd/aws/aws_summary.rs:268`):

```rust
let abused_aws_api_calls =
    read_abused_aws_api_calls(&resolve_rules_config_path("rules/config/abused_aws_api_calls.csv"));
```

Signature and warning (`src/cmd/aws/aws_summary.rs:636`, `652-659`):

```rust
fn read_abused_aws_api_calls(file_path: &Path) -> HashMap<String, String> {
    // ...unchanged Ok(...) arm...
        Err(_) => {
            p(
                Some(Color::Rgb(255, 0, 0)),
                &format!(
                    "Warning: could not open the abused AWS API calls file ({}). \
                     All API calls will be classified as non-abused.",
                    file_path.display()
                ),
                true,
            );
            HashMap::new()
        }
}
```

## Verification

Add a unit test in the existing `#[cfg(test)]` module of `src/cmd/aws/aws_summary.rs` covering the reader's degraded path and the resolver fallback:

```rust
#[test]
fn missing_abused_csv_returns_empty_map() {
    // A path that does not exist must yield an empty map (not a panic),
    // which is what drives the "all non-abused" degradation we warn about.
    let map = read_abused_aws_api_calls(std::path::Path::new(
        "definitely/does/not/exist/abused_aws_api_calls.csv",
    ));
    assert!(map.is_empty());
}

#[test]
fn resolve_rules_config_prefers_existing_cwd_path() {
    let dir = TempDir::new().unwrap();
    let rel = dir.path().join("abused.csv");
    std::fs::write(&rel, "eventName,description\nRunInstances,launch\n").unwrap();
    let resolved = resolve_rules_config_path(rel.to_str().unwrap());
    assert!(resolved.exists());
    let map = read_abused_aws_api_calls(&resolved);
    assert_eq!(map.get("RunInstances").map(String::as_str), Some("launch"));
}
```

For the `--regex` / `--filter` fatal paths (which call `std::process::exit` and so cannot be asserted in-process), verify manually:

```
# clean error, no backtrace, exit code 1:
$ suzaku aws-ct-search -d ./logs --regex '(unclosed'
Invalid --regex pattern: regex parse error: ...

# clean error on a malformed filter instead of a silently-ignored filter:
$ suzaku aws-ct-search -d ./logs --filter 'eventName=ConsoleLogin'
Invalid --filter value 'eventName=ConsoleLogin'. Expected FIELD:VALUE format.

# abused classification now works from any cwd:
$ cd /tmp && suzaku aws-ct-summary -d /path/to/logs -o out.csv
# out.csv abused-API sections are populated
```

---

Shirofune-Security:
Fixes #162.

## Problem & fix
Three `aws-ct-*` input-handling gaps:

1. **`--regex` panicked on an invalid pattern.** `aws-ct-search` compiled it with `Regex::new(pattern).expect("Invalid regex pattern")` (`aws_search.rs`), so a bad pattern aborted with a Rust panic + backtrace. → Now compiled with a `match` that routes the error through `fatal_error` for a clean message and exit.

2. **A malformed `--filter` was silently dropped.** `parse_filter_conditions` used `if let Some((field, value)) = filter.split_once(':')`, so a filter with no `:` was silently ignored — the search then ran with that condition missing, giving misleadingly broad results. → The call site now validates each `--filter` up front and rejects a colon-less value with a clear `expected FIELD:VALUE` error (the `parse_filter_conditions` signature and its 3 tests are unchanged).

3. **abused-API classification message was unhelpful** (correcting the issue's wording — it was **not** silent). `read_abused_aws_api_calls` already printed a red warning on failure, but named neither the path nor the impact. → The warning now reports the path it looked for and states that all API calls will be classified as non-abused. (Resolving the path relative to the executable/`rules` dir is left as a follow-up; this makes the existing warning actionable.)

## Testing
`cargo check` clean; `cargo test parse_filter` (4 tests) passes. The regex/filter error paths call `fatal_error` (process exit), verified by build + manual check.


Shirofune-Security:
CI is failing in `cargo fmt -- --check`. The reported diff is in `src/cmd/aws/aws_search.rs` around the new invalid-regex `fatal_error(...)` call. Please run `cargo fmt` and push the formatted file.

fukusuket (approved):
LGTM!!🚀

PR #170, opened 2026-07-20
fix: warn instead of silently skipping unreadable input files (#161)

### LINKED ISSUE #161: Scan silently loses data: unreadable files and unparseable events are skipped with no warning, and plain (non-gz) reads are unbounded
Suzaku reports `Total log files: N` up front (from `count_files_recursive`), but three separate spots in `src/core/scan.rs` can silently drop inputs from the actual analysis without warning or accounting, so the number the operator sees can overstate what was really scanned. A file that can't be read (permission denied, invalid UTF-8, corrupt gzip) is counted but skipped at read time with `continue`; an event that fails JSON re-parse in `detect_events` is dropped and is not included in `total_events`; and the recent gzip-bomb bound (#147) only guards the `.gz` path, leaving the plain `.json`/`.jsonl`/`.csv` path doing an unbounded whole-file `fs::read_to_string` that can still OOM the process. Because this is a forensic tool, silent input loss undermines the core guarantee that "no hits" means "nothing matched" rather than "some data was never looked at."

## Problem

**(1) Unreadable files are counted but silently skipped — `src/core/scan.rs:179-187`.** In the main scan loop the plain-text read swallows every error and `continue`s, incrementing only the progress bar:

```rust
let log_contents = if path_str.ends_with("json")
    || path_str.ends_with("jsonl")
    || path_str.ends_with("csv")
{
    match fs::read_to_string(&path) {
        Ok(contents) => contents,
        Err(_) => {
            if show_progress {
                pb.inc(1);
            }
            continue;
        }
    }
} else if path_str.ends_with("gz") {
    match read_gz_file(&path) {
        Ok(contents) => contents,
        Err(_) => {
            if show_progress {
                pb.inc(1);
            }
            continue;
        }
    }
} else {
```

That file was already tallied by `count_files_recursive` (`src/core/scan.rs:585`) and printed as part of `Total log files:` (`src/core/scan.rs:133-134`), so the count now overstates what was analyzed, with nothing on stderr to say a file was dropped. (The convenience helper `get_content` at `src/core/scan.rs:699-708` has the same `unwrap_or_default()` behavior.)

**(2) Unparseable events are silently dropped and not counted — `src/core/scan.rs:442-445`, `469`.** In `detect_events`, each event that passes the time filter is re-serialized and re-parsed via `event_from_json`; a parse failure is discarded, and `total_events` is then incremented only by the count of *successfully parsed* events:

```rust
.filter_map(|event| match event_from_json(event.to_string().as_str()) {
    Ok(json_event) => Some((event, json_event)),
    Err(_) => None,
})
.collect();
```

```rust
summary.total_events += json_events.len();
```

So an event that survived ingestion but fails re-parse is invisible: it is never rule-matched and never appears in any statistic, meaning a real detection could be missed with zero diagnostic output.

**(3) The #147 bound only covers gzip — `src/core/scan.rs:179`, `622-651`.** The decompression-bomb fix caps the `.gz` path via `read_gz_file` / `MAX_DECOMPRESSED_BYTES` (3 GiB). The plain path at `src/core/scan.rs:179` still calls `fs::read_to_string(&path)`, which reads the entire file into a `String` in one shot with no ceiling. A single very large non-gz input (a multi-GB exported `.jsonl`, or a sparse/crafted file) can therefore still exhaust memory and OOM-kill the whole scan — the exact failure mode #147 set out to prevent, just through the sibling code path.

## Impact

Suzaku is a locally-run forensic timeline / threat-hunting CLI, so this is a data-integrity and robustness issue rather than a remote-exploitation one, but the values involved (CloudTrail / Azure / M365 records, and the file set an analyst points it at) are attacker-influenceable and are exactly what an incident responder is trying to reason about:

- **Silent under-reporting of coverage (1, 2).** An analyst who sees `Total log files: 5,000` and no detections may conclude the environment is clean, when in fact some files were unreadable or some events failed re-parse and were never matched. In IR, a false "nothing found" is worse than a loud error. An attacker who can plant one malformed record (invalid UTF-8, a subtly non-conforming JSON object) in an otherwise valid batch can cause the *malicious* event to be the one dropped, while everything around it is analyzed normally.
- **Denial of service via a single large file (3).** Feeding one oversized non-gz log can OOM the process and abort the entire run, not just that file.

**Concrete repro:**

- *(1)* In a scan directory, `chmod 000 one.json` (or write a `.json` file containing invalid UTF-8 bytes). Run `suzaku` against the directory: the header still counts that file, the scan completes, and no message indicates it was skipped.
- *(2)* Include an event that ingests but round-trips to something `event_from_json` rejects; it never matches a rule and is absent from `total_events`, with no warning.
- *(3)* `fallocate -l 8G huge.jsonl` in the scan directory and run the scan on a machine with < 8 GB free — the plain read buffers the whole file and the process is OOM-killed, whereas the same-sized `.gz` would have been bounded and skipped with a warning.

## Suggested fix

Make ingestion loss visible and bound plain reads, mirroring the existing `read_gz_file_capped` style.

**Bound + report plain reads.** Add a capped plain reader next to the gz one (reusing `MAX_DECOMPRESSED_BYTES`), and have the loop warn on any skip:

```rust
/// Reads a plain (uncompressed) log file, refusing to buffer more than `max_bytes`.
/// Mirrors `read_gz_file_capped` so the plain `.json`/`.jsonl`/`.csv` path is bounded
/// the same way the `.gz` path is (issue #147 covered only gzip).
fn read_plain_file_capped(file_path: &PathBuf, max_bytes: u64) -> io::Result<String> {
    let size = fs::metadata(file_path)?.len();
    if size > max_bytes {
        eprintln!(
            "[WARNING] Skipping {}: file size exceeds the {} GiB limit.",
            file_path.display(),
            max_bytes / (1024 * 1024 * 1024)
        );
        return Err(io::Error::new(
            io::ErrorKind::InvalidData,
            "file size exceeds the maximum allowed limit",
        ));
    }
    fs::read_to_string(file_path)
}

pub fn read_plain_file(file_path: &PathBuf) -> io::Result<String> {
    read_plain_file_capped(file_path, MAX_DECOMPRESSED_BYTES)
}
```

Then in the scan loop (`src/core/scan.rs:175-203`), route plain reads through it and warn + count every skip. Declare `let mut skipped_files: u64 = 0;` before the `for path in file_paths` loop:

```rust
let log_contents = if path_str.ends_with("json")
    || path_str.ends_with("jsonl")
    || path_str.ends_with("csv")
{
    match read_plain_file(&path) {
        Ok(contents) => contents,
        Err(e) => {
            eprintln!("[WARNING] Skipping {}: {e}", path.display());
            skipped_files += 1;
            if show_progress {
                pb.inc(1);
            }
            continue;
        }
    }
} else if path_str.ends_with("gz") {
    match read_gz_file(&path) {
        Ok(contents) => contents,
        Err(e) => {
            eprintln!("[WARNING] Skipping {}: {e}", path.display());
            skipped_files += 1;
            if show_progress {
                pb.inc(1);
            }
            continue;
        }
    }
} else {
```

After the loop (before `Ok(())`), surface the total:

```rust
if skipped_files > 0 {
    eprintln!(
        "[WARNING] {skipped_files} of {count} log file(s) could not be read and were not analyzed.",
    );
}
```

**Count dropped events.** In `detect_events`, aggregate re-parse failures and report them once. Declare an atomic before the `for event_chunks` loop and read it after:

```rust
let dropped_events = std::sync::atomic::AtomicUsize::new(0);
for event_chunks in events.chunks(CHUNK_SIZE) {
    // ...
    .filter_map(|event| match event_from_json(event.to_string().as_str()) {
        Ok(json_event) => Some((event, json_event)),
        Err(_) => {
            dropped_events.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
            None
        }
    })
    .collect();
    // ...
}
let dropped = dropped_events.load(std::sync::atomic::Ordering::Relaxed);
if dropped > 0 {
    eprintln!(
        "[WARNING] {dropped} event(s) could not be re-parsed as JSON and were not analyzed.",
    );
}
```

Checklist:

- [ ] (1) Warn on unreadable files and track a skipped-file count reported at end of scan.
- [ ] (2) Count and warn about events dropped at `event_from_json` re-parse.
- [ ] (3) Cap plain `.json`/`.jsonl`/`.csv` reads the same way `.gz` is capped, warning + skipping oversized files.

## Verification

Add unit tests alongside the existing `read_gz_file_capped_*` tests in `#[cfg(test)] mod tests` (they already use `tempfile::tempdir()`):

```rust
#[test]
fn read_plain_file_capped_rejects_oversized_file() {
    // A file larger than the cap must error (so the caller skips it) rather than
    // buffering the whole thing into memory.
    let dir = tempfile::tempdir().unwrap();
    let path = dir.path().join("big.json");
    fs::write(&path, "[]").unwrap();
    assert!(read_plain_file_capped(&path, 1).is_err());
}

#[test]
fn read_plain_file_capped_accepts_within_limit() {
    let dir = tempfile::tempdir().unwrap();
    let path = dir.path().join("ok.json");
    fs::write(&path, "[]").unwrap();
    assert_eq!(read_plain_file_capped(&path, 1024).unwrap(), "[]");
}
```

Run `cargo test read_plain_file_capped`.

Manual check for the warnings: create a scan directory with (a) a `.json` file `chmod 000`, and (b) an event that fails `event_from_json`, then run a scan and confirm stderr now emits `[WARNING] Skipping ...`, the end-of-scan `X of N log file(s) could not be read ...` line, and the `N event(s) could not be re-parsed ...` line — none of which appear on current `main`.

---

Shirofune-Security:
Fixes #161 (the silent-skip part; see scope notes).

## Problem
When scanning, a file that was **counted** in the file total but then failed to read was dropped **silently**:
- `process_events_from_dir` (`scan.rs`) matched `fs::read_to_string`/`read_gz_file` and did a bare `continue` on `Err(_)`.
- `get_content` (single-file input) returned `""` via `unwrap_or_default()` on any read error.

So a permission-denied file, a non-UTF-8-content file, a file removed mid-scan, or a corrupt `.gz` was skipped with no indication — the reported "Total log files" overstated what was actually analyzed.

## Fix
Emit `[WARNING] Skipping <file>: <reason>` at both the directory-scan `Err` arms and in `get_content`. The gzip size-cap warning (from #147) was moved from inside `read_gz_file_capped` into its returned error, so it now surfaces through the **same single** warning path (no double-warning for the bomb case, and corrupt/oversized `.gz` are now reported consistently).

## Scope notes (deliberately not done here)
- **Plain reads are not size-capped like `.gz`.** The #147 cap exists because gzip *amplifies* (a few MB → many GB). A plain `.json`/`.jsonl`/`.csv` is bounded by its on-disk size, so capping it would reject legitimate large CloudTrail exports — a regression, not a fix.
- The rare `event_from_json` drop inside `detect_events` (a Rayon `par_iter().filter_map`) is left as-is; counting/warning there needs cross-thread plumbing and the input is already-validated `Value` re-parsed, so failures are edge cases. Noted for a possible follow-up.

## Testing
`cargo check` clean; `cargo test read_gz_file` (3 tests) passes (the size-cap error path still returns `Err`, now with a descriptive message). The change adds stderr warnings; return-value behavior is unchanged.


fukusuket (approved):
LGTM!!🚀

Issue #159, opened 2026-07-20
Azure/M365 GeoIP enrichment is a silent no-op (source-IP field hardcoded to `sourceIPAddress`)


Shirofune-Security:
When `--geo-ip`/`-G` is used with `azure-timeline` (or `m365-timeline`), the `SrcASN`, `SrcCity`, and `SrcCountry` columns are always emitted as `-`, even for events that carry a routable public IP. The geo-enrichment branch in `get_value_from_event_common` reads the source IP from a hardcoded `event.get("sourceIPAddress")`, which is the AWS CloudTrail field name. The Azure/M365 profile stores the caller IP under different keys (`callerIpAddress`, `ClientIP`, `ActorIpAddress`, `claims.ipaddr`), so the lookup never finds a value and the code falls through to the `-` placeholder. The geo columns are still injected into the profile for Azure, so the output silently claims "no geolocation available" for every record. AWS is unaffected because its profile happens to map `SrcIP` to `.sourceIPAddress`, which matches the literal.

## Problem

`get_value_from_event_common` (`src/core/timeline_writer.rs:385`) resolves the source IP for the three geo columns from a hardcoded field name:

```rust
if matches!(key, "SrcASN" | "SrcCity" | "SrcCountry") {
    if let Some(geo) = geo_ip
        && let Some(ip) = event.get("sourceIPAddress")
    {
        let ip = ip.value_to_string();
        if let Some(ip) = geo.convert(ip.as_str()) {
            return match key {
                "SrcASN" => geo.get_asn(ip),
                "SrcCity" => geo.get_city(ip),
                _ => geo.get_country(ip),
            };
        }
    }
    return "-".to_string();
}
```

(hardcoded key at `src/core/timeline_writer.rs:398`, fall-through placeholder at `:409`.)

The literal `"sourceIPAddress"` matches the AWS profile (`config/aws_profile.yaml:10` → `SrcIP: '.sourceIPAddress'`) but not the Azure/M365 profile (`config/azure_profile.yaml:9`):

```yaml
SrcIP: '.claims.ipaddr|.callerIpAddress|.ClientIP|.ActorIpAddress'
```

None of those keys is named `sourceIPAddress`, so `event.get("sourceIPAddress")` returns `None` and the branch returns `"-"`.

Meanwhile `load_profile` (`src/core/util.rs:135`) still injects the three geo columns whenever geo lookup is enabled, regardless of log source (`src/core/util.rs:154-158`):

```rust
[...]

PR #171, opened 2026-07-20
fix: resolve GeoIP source IP from the profile SrcIP spec (#159)

Shirofune-Security:
Addresses #159 (the Azure/M365 no-op; see note on the two minor items).

## Problem
`get_value_from_event_common` (`src/core/timeline_writer.rs`) resolved the source IP for the `SrcASN`/`SrcCity`/`SrcCountry` columns from a hardcoded `event.get("sourceIPAddress")` — an AWS-CloudTrail-only field. `config/aws_profile.yaml` maps `SrcIP: '.sourceIPAddress'` (matches, so AWS works), but `config/azure_profile.yaml` maps `SrcIP: '.claims.ipaddr|.callerIpAddress|.ClientIP|.ActorIpAddress'`. So for `azure-timeline`/M365, `event.get("sourceIPAddress")` was always `None` and the three geo columns rendered `-` for **every** event — a silent dead feature even when the record carried a routable public IP. The same hardcode also affected the correlation path.

## Fix
- Added `src_ip_spec(profile)` returning the profile's declared `SrcIP` field spec.
- Threaded `src_ip: &str` through `get_value_from_event` / `get_value_from_event_common` / `get_value_from_correlation_event` and their call sites (`write_record`, `build_correlation_record`, the two raw-output paths).
- The geo branch now resolves the IP by walking the spec's `|`-separated keys with `event.get()` — the same fallback already used for `.`-prefixed columns (which is why the `SrcIP` column itself already resolved for Azure). **AWS behavior is unchanged** (its spec is `.sourceIPAddress`).

## Testing
`cargo test` → **122 passed**. New tests: `src_ip_spec_reads_profile_srcip_field` (helper), and `geoip_resolves_source_ip_via_profile_spec` — asserts the same IP under the AWS `sourceIPAddress` field and the Azure `callerIpAddress` field enrich **identically** (this fails before the fix, since the Azure field was ignored). The existing #145 regression test still passes.

## Note on the two minor GeoIP cleanups in #159
The issue also listed two `geoip.rs` low-priority items — `convert()` mishandling hex-form IPv4-mapped IPv6, and `get_asn` not caching lookup errors. Those are a separate concern from this fix and are left as follow-up, so I've marked this **"Addresses"** rather than "Fixes" to keep #159 open until they're done.


Shirofune-Security:
CI is failing in `cargo fmt -- --check`. The reported diffs are in `src/core/timeline_writer.rs` around the new `get_value_from_event(..., src_ip_spec(profile))` calls and the long test assertions/event constructors. Please run `cargo fmt` and push the formatted result.

fukusuket (approved):
LGTM!!🚀

PR #172, opened 2026-07-20
feat: add DuckDB output and name-based -t format selection

Shirofune-Security:
## What

Adds **DuckDB output** for `aws-ct-timeline` and `azure-timeline`, and reworks `-t, --output-type` to select formats by **name** so any combination can be written at once.

```
suzaku aws-ct-timeline -d ../suzaku-sample-data -o result -t csv,jsonl,duckdb
# -> result.csv + result.jsonl + result.duckdb (all in one scan)
```

## Details

- **Named formats.** `-t` now takes `csv` (default), `json`, `jsonl`, `duckdb` as a comma-separated / repeatable list (clap `ValueEnum`) instead of the old numeric `1..5`. Repeats are de-duplicated, and each requested format writes its own file (extension auto-corrected from the `-o` path).
- **DuckDB writer.** Output is a self-contained `.duckdb` database with a `timeline` table whose columns are the output-profile headers (all `VARCHAR`). Rows are inserted as the scan streams, so it stays memory-friendly. The DuckDB engine is compiled in via the `bundled` feature — no system `libduckdb` is needed, keeping releases a single binary.
- `init_writers` now takes the format list plus the profile (needed for the DuckDB table schema) and fans out one sink per requested format.
- The `--raw-output` guard now checks for `json`/`jsonl` in the format list.
- Applies to the commands sharing `OutputOption`: `aws-ct-timeline`, `azure-timeline`, and `aws-ct-search`. `aws-ct-summary` keeps its own separate `-t`.

## Breaking change

The numeric `-t 1..5` form is removed in favor of format names. Since 2.0.0 is unreleased this is a good time to make the switch.

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, and `cargo test` all pass (added a `timeline` sink unit test covering table creation, a hyphenated column name, and row inserts).
- End-to-end against `suzaku-sample-data`: `-t csv,jsonl,duckdb` produced all three files; the `.duckdb` `timeline` table row count and columns matched the CSV exactly. `-t parquet` is cleanly rejected; `-t csv,csv` de-dups.

Docs (EN command reference) and CHANGELOG (EN/JA) updated.


Shirofune-Security:
I found two things worth fixing before merge:

1. The `--clobber` preflight still checks only the literal `-o` path in `src/main.rs`, but `init_writers()` writes to normalized/generated paths such as `<base>.duckdb`, `<base>.json`, and `<base>.jsonl`. With the new DuckDB output this can overwrite an existing database without `-C`, e.g. `-o result -t duckdb` will replace the `timeline` table in `result.duckdb` even when `result` itself does not exist. Please precompute/check the actual output paths for the selected formats before opening writers, similar to the summary command tests that preflight all output paths.

2. In `OutputContext::flush_all()`, the no-hit cleanup closes/removes CSV/JSON/JSONL writers before deleting files, but it leaves `self.writers.duckdb` alive. On Windows the open DuckDB connection can keep the `.duckdb` file locked, so `remove_file()` will fail silently and leave an empty database behind even though `Results saved: None` is reported. Please drop/clear the DuckDB sink before removing output files when `!self.has_written`.

Local checks passed for me with the Rust env loaded: `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --locked` (121 tests).

Shirofune-Security:
Both addressed in the latest push:

1. **`--clobber` now preflights the real output paths.** Extracted the base→`<base>.<ext>` resolution into a shared `resolve_output_paths()` (single source of truth with `init_writers`), and `main.rs` now checks every file the selected formats would write, erroring and naming the ones that already exist. Verified: `-o result -t duckdb` with a pre-existing `result.duckdb` (and no `result`) is now refused without `-C`, and `-C` overrides as before.

2. **Empty DuckDB no longer left behind.** `flush_all()`'s no-hit branch now drops the DuckDB sink (closing the connection so the `.duckdb` file isn't locked on Windows) before `remove_file()`. While testing this I found `aws-ct-search` never called `flush_all()` at all, so it left an empty file behind for *every* format despite reporting `Results saved: None` — added the `flush_all()` call there too. Verified end-to-end: a no-match search with `-t csv,duckdb,jsonl` now removes all three; a matching run still writes them (310 CSV rows == 310 DuckDB `timeline` rows).

Added regression tests (`resolve_output_paths_expands_base_per_format_and_dedups`, `flush_all_drops_duckdb_sink_and_removes_empty_db_on_no_hits`). `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, and `cargo test` (123) all green.

YamatoSecurity (approved):

PR #176, opened 2026-07-26
fix: write warnings/errors to a log file and unify run-ending errors in red

fukusuket:
## Summary

Warnings and errors are no longer printed to the terminal. They are written to
`logs/errorlog-<YYYYMMDD_HHMMSS>.log`, with the command line that produced them on the
first line. Separately, every error that ends the run is now printed in red and honors
`--no-color` — previously the color was inconsistent across call sites.

## Motivation

Per-file warnings (`[WARNING] Skipping <file>: <reason>`, added for the "silently skipped
files" fix) were written to stderr while the progress bar was live. The bar redraws by
moving the cursor up over however many lines it wrote last time, so any write it does not
know about shifts the cursor and strands the previous block on screen — the bar visibly
appears twice. Scanning a directory with many unreadable files also buried the results
summary under hundreds of lines.

Sending these messages to a file fixes both problems and leaves an artifact that can be
attached to a bug report.

### 1. Error log file (`src/core/errorlog.rs`, new)

- `log_warn()` / `log_error()` append to `logs/errorlog-<YYYYMMDD_HHMMSS>.log`.
- The file (and the `logs/` directory) is created **lazily on the first message**, so a
  clean run leaves nothing behind. `logs/` is already in `.gitignore`.
- The **first line is the command line** that produced the log, quoted so it can be pasted
  back into a shell; then a blank line, then the messages.
- Each message is flushed immediately: `fatal_error()` calls `process::exit`, which does
  not run destructors, so a buffered writer would lose everything.
- If `logs/` cannot be written (read-only mount, permissions), messages fall back to stderr
  instead of being dropped.

Moved to the log file (diagnostics raised while processing logs):

| Message | Location |
| --- | --- |
| `Skipping <file>: <reason>` (×3) | `core/scan.rs` |
| `Failed to scan directory <dir>: <e>` | `core/scan.rs`, `cmd/aws/aws_metrics.rs`, `cmd/aws/aws_summary.rs` |
| `Error processing correlation events: <e>` | `core/timeline.rs` |
| `Could not open the abused-AWS-API list at ...` | `cmd/aws/aws_summary.rs` |

`fatal_error()` is the one exception: it is written to the log **and** kept on the terminal,
since the process exits right after and a file-only message would look like a silent exit.

Pre-flight CLI validation (bad path, invalid `--min-level`, existing output without `-C`,
missing rules/profile, `update-rules` failures) stays on the terminal as immediate
feedback — logging it would create a `logs/` directory for a simple typo.

At the end of a run, a single pointer line is printed only when something was recorded:

```
Warnings and errors: 2 messages saved to logs/errorlog-20260726_101207.log
```

Example log file:

```
suzaku aws-ct-timeline -d ./ct -o out.csv -q

[WARNING] Skipping ./ct/bad file.json: Permission denied (os error 13)
[WARNING] Skipping ./ct/broken.gz: unexpected end of file
```

### 2. Consistent red for run-ending errors

Added a single `error_msg(no_color, msg)` helper in `core/util.rs`; `fatal_error()` now
goes through it, so the color of a terminal error is decided in one place.

- **Default terminal color → red:** the four `check_path_exists()` messages (they used bare
  `println!`, so they had no color at all), `The output file already exists: ...` (×2),
  `Rule file or directory does not exist`, `--raw-output option is only available in JSON
  formats`, `Invalid minimum level`, `Profile file does not exist` (aws/azure), and the six
  `update-rules` failure messages.
- **Hard-coded red that ignored `--no-color` → `Red.rdg(no_color)`:** `No events found.`
  (summary + metrics) and `The file ... already exists. Use --clobber to overwrite.`

`no_color` is now threaded through `check_path_exists()`, `validate_min_level()`,
`update_rules()`, `pull_repository()` and `clone_rules()`.

## Testing

- `cargo test` — 136 passed, 0 failed (5 new tests in `core::errorlog` covering the
  command-line header, the `errorlog-<timestamp>.log` filename, directory creation, and
  argument quoting/escaping).
- `cargo fmt --check` and `cargo clippy --all-targets` are clean.
- Manual run against a directory containing a `chmod 000` JSON file and a corrupt `.gz`:
  both warnings landed in the log file with the command-line header, nothing leaked to the
  terminal, the progress bar rendered correctly, and the pointer line appeared at the end.
- Manual run of each run-ending error: all emit `\e[38;2;255;0;0m` (red) and drop the color
  under `--no-color`. The `update-rules` failure paths were not exercised (they need a git
  failure), but they use the same `error_msg()` helper.

## Notes

- No new dependencies; no CLI option changes.
- The log location is fixed at `./logs` relative to the working directory, matching Hayabusa.


fukusuket:
### before
```
Scanning now. Please wait.

[00:00:39] 127,540 / 340,656 ⠠ [==============>                         ] 37%                                                                                                                                                                                                                                       
../data/aws/CloudTrail/ap-northeast-1/2024/03/21/000123456789_CloudTrail_ap-northeast-1_20240321T1525Z_SnckBmIntS2bUVl4.json.gz (669 B)
[00:02:01] 340,656 / 340,656   [========================================] 100%
```

### after
```
Scanning now. Please wait.

[00:02:03] 340,656 / 340,656   [========================================] 100%                                                                                                                                                                                                                                      Scanning finished.
```

```
% cat logs/errorlog-20260726_104248.log
./suzaku aws-ct-timeline -d ../data/aws/CloudTrail -G ../data/maxmind -t csv,json -o sample-aws-ct-timeline-2 -C -q

[WARNING] Skipping ../data/aws/CloudTrail/ap-northeast-1/2024/03/21/000123456789_CloudTrail_ap-northeast-1_20240321T0000Z_ZASADRphLTxGGOGw.json.gz: unexpected end of file
```

YamatoSecurity (approved):
@fukusuket LGTM! Thanks for fixing!

PR #177, opened 2026-07-26
feat: expand `aws-ct-metrics` into a multi-field triage command

fukusuket:
## What

`aws-ct-metrics` can now aggregate **several fields in a single scan**, reports a time window per value, enriches IP addresses with GeoIP, and writes the same output formats as the rest of the tool. It also fixes two long-standing warts in the command: the hardcoded `EventName` column header and the silent overwriting of `-o` files.

```bash
# One scan, five fields, GeoIP on the source IPs, CSV + DuckDB out
./suzaku aws-ct-metrics -d ../suzaku-sample-data \
  -F sourceIPAddress,userAgent,userIdentity.arn,awsRegion,userIdentity.accessKeyId \
  -G ../GeoLite2-DBs -o metrics -t csv,duckdb
```

## Why

The five fields this PR was asked to add (`sourceIPAddress`, `userAgent`, `userIdentity.arn`, `awsRegion`, `userIdentity.accessKeyId`) were already reachable through `-F`, including the nested ones in dot notation. What made them impractical in real investigations was everything around the aggregation:

| Problem | Effect on an investigation |
| --- | --- |
| One field per run | Five fields meant five full scans of the same S3 export |
| Column header hardcoded to `EventName` | A run over `sourceIPAddress` produced a column of IPs headed `EventName` |
| No GeoIP | Source IPs had to be enriched by hand or via `aws-ct-summary` |
| No first/last seen | No way to tell a value that appeared once from one that spanned the whole dataset |
| CSV only, no `-t`, no `-C` | Inconsistent with `aws-ct-timeline` / `aws-ct-summary`; `-o` silently clobbered files |
| Events missing the field were dropped | Percentages were a share of "events that had this field", which is not what the header implied |
| No validation of `-F` | A mistyped field name scanned the whole dataset and returned nothing useful |

So this is deliberately framed as **finishing the existing `-F` option**, not as a new aggregation axis.

### Aggregation

- `-F, --field-name` now takes a comma-separated list (or repeats). Every field is tallied in **one pass** over the logs.
- Each value carries `Count`, `Percent`, `FirstSeen` and `LastSeen`. The per-value time window uses the same `upsert_count_entry` helper as `aws-ct-summary`, which moved to `core/util.rs` so both commands cannot drift apart.
- Events **missing** the requested field are counted as `-` instead of being dropped, so a field's total always equals the number of events scanned and `Percent` means "share of all events". This is also informative in itself: the `-` share of `userIdentity.accessKeyId` is the share of activity that came from console sessions rather than access keys.
- **`-F` is validated before the scan starts** (`src/option/ct_fields.rs`, wired in as a clap `value_parser`). Counting missing values as `-` means a mis-cased name would otherwise produce a plausible-looking table of one `-` row at 100% — after a full pass over the dataset. Now:

  ```
  $ suzaku aws-ct-metrics -d logs -F sourceIPaddress
  error: invalid value 'sourceIPaddress' for '--field-name <FIELD_NAME,...>':
         'sourceIPaddress' is not a CloudTrail field. Did you mean 'sourceIPAddress'? (field names are case-sensitive)
  ```

  The list holds the top-level CloudTrail record fields plus the documented paths inside `userIdentity` and `tlsDetails`. Containers whose contents are API-specific (`requestParameters`, `responseElements`, `additionalEventData`, `serviceEventDetails`, `insightDetails`, `resources`, `addendum`, `edgeDeviceDetails`) accept **any** sub-path, so `-F requestParameters.bucketName` keeps working. The suggestion is case-insensitive only — no fuzzy matching, since silently aggregating a field the user did not ask for is worse than the error.
- A validly named field that these particular logs do not carry (`errorCode` in an export with no failures) still prints a one-line notice, since a lone `-` row at 100% reads like a real result.
- `-s, --include-sts-keys` mirrors `aws-ct-summary`: temporary `ASIA...` STS keys are excluded by default because a fresh key per session inflates the value count.

### Output

- `-G, --geo-ip` adds `SrcASN` / `SrcCity` / `SrcCountry`. Unlike `aws-ct-summary`, which folds the enrichment **into the aggregation key** (`1.2.3.4 (ASN, City, Country)`), the key here stays the bare IP and the enrichment goes into its own columns — so the result can be re-aggregated by ASN or country, and a GeoIP database update cannot split one IP's counts across two rows. Only values that parse as an IP are enriched; an AWS-service caller such as `cloudtrail.amazonaws.com` shows `-`.
- `-t, --output-type csv,json,jsonl,duckdb` and `-C, --clobber`, resolved through the same `resolve_output_targets()` the timeline and summary commands use. The clobber check runs **before** the GeoIP lookups, so an existing file fails immediately.
- DuckDB output is a single `metrics` table. Unlike the summary, there is no nesting to preserve — one row per (field, value) is already the relational shape, so `SELECT * FROM metrics WHERE Field = 'sourceIPAddress' ORDER BY Count` works directly.
- Values are ordered by count, most frequent first, within each field. Rows with equal counts are sorted by value, so two runs over the same data produce byte-identical files instead of shuffling in `HashMap` order.

### Screen vs. file layout

Screen output prints one table per field, with the **first column named after the field** — which is what the hardcoded `EventName` header was always meant to be:

```
+-----------------+-------+---------+---------------------+---------------------+--------+---------+------------+
| sourceIPAddress | Count | Percent |      FirstSeen      |       LastSeen      | SrcASN | SrcCity | SrcCountry |
+===============================================================================================================+
| 81.2.69.142     | 1,024 | 62.19%  | 2021-07-05 13:03:12 | 2021-07-05 13:03:50 | ...    | London  | UK         |
```

File output is one flat table with a `Field` column, so every field lands in a single CSV/JSON/DuckDB file.

## Breaking changes

- **CSV columns** changed from `EventName,Percent,Total` to `Field,Value,Count,Percent,FirstSeen,LastSeen` (+ `SrcASN,SrcCity,SrcCountry` with `-G`). Anything parsing the old CSV needs updating.
- **`-o` no longer overwrites silently.** Existing scripts that re-run into the same path now need `-C`.
- `Percent` for `eventName` shifts slightly on datasets containing events without an `eventName`, since those are now counted as `-` rather than excluded from the denominator.

## Relationship to `aws-ct-summary`

The two commands overlap in *fields* but not in *axis*, and this PR leans into that rather than blurring it:

- `aws-ct-metrics` aggregates **across all events** — the overall distribution, one row per value, ready to sort or filter in a spreadsheet or in SQL. The entry point for triaging a dataset.
- `aws-ct-summary` breaks the same information down **per user ARN** — one row per principal, with multi-line cells. The tool for investigating a specific principal.

The summary's CSV packs a principal's IPs into a single multi-line cell, which is right for reading one principal but useless for asking "which source IPs exist at all?". The reverse is equally true. To keep the choice obvious for users, this PR:

- adds a **See also** note in both directions in the docs (`analysis.md` ↔ `dfir-summary.md`, EN + JA), stating the axis in one line;
- aligns the option surface — `-s`, `-C`, `-G`, `-t`, `-o` now mean the same thing, with the same short flags, in both commands;
- deliberately does **not** add global aggregation to `aws-ct-summary`, which would break its one-row-per-principal contract.

## Testing

- 11 new unit tests in `aws_metrics.rs` covering multi-field aggregation in one pass, per-value first/last seen, missing fields counted as `-`, the warning for a field no event has (and its absence for a partially present field), STS exclusion with and without `-s`, count ordering within each field, GeoIP enrichment (asserted against the `test_files/mmdb` fixtures: `81.2.69.142` → London / United Kingdom) and its absence without `-G`, CSV header shape, and deterministic tie-breaking.
- 4 new unit tests in `ct_fields.rs` covering the documented fields, open-container sub-paths, the mis-cased suggestion, and rejection of an unknown name (including an unknown container with a dotted path).
- Full suite: `cargo test` 152 passed, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt` applied.
- Manually verified against `test_files/json`: multi-field screen output, directory scan, all four output formats written together, the `-C` guard firing on an existing file, and the DuckDB `metrics` table queried with the `duckdb` CLI.

## Documentation

- `website/docs/commands/analysis.md` / `analysis.ja.md`: new prose on multi-field aggregation, the `-` semantics, the STS default, the axis difference vs. `aws-ct-summary`, four new examples, and an output-layout section.
- `website/docs/commands/dfir-summary.md` / `.ja.md`: See-also pointer back to `aws-ct-metrics`.
- The usage block was refreshed in all 15 locales; the surrounding prose and examples in the other 13 languages still need translating (they currently describe the pre-existing behaviour, which remains accurate).
- `CHANGELOG.md` / `CHANGELOG-Japanese.md`: one Enhancement and one Bug Fix entry. `website/docs/resources/changelog*.md` was left alone since it already lags the root changelog and appears to be synced in batches.
- `OLD-README*.md` were intentionally not touched.

## Follow-ups (not in this PR)

- `errorCode` is probably the single highest-value field to aggregate next (`AccessDenied` spikes = permission enumeration); it works today via `-F errorCode` and needs no code change.
- Cross-tabulation (`--cross eventName,errorCode`) and unique-principal counts per value (`--unique-by`) were considered and left out to keep this reviewable.
- High-cardinality fields: unique values are held in a `HashMap` for the whole scan and every one of them is written out. On a large tenant a `userAgent` aggregation produces a very long table, and the memory grows with the number of distinct values. Row limiting and alternative sort orders were considered and deliberately left out of this PR.


YamatoSecurity:
@fukusuket Do empty values produce `-` in the output? It might be better to just output an empty string to save space. What do you think?

fukusuket:
@YamatoSecurity 
Thank you for the suggestion! I've updated the code to output an empty string!

YamatoSecurity (approved):
LGTM! Thank you!! @fukusuket

PR #180, opened 2026-07-27
feat: Make the DuckDB output a typed, self-describing data interface

fukusuket:
## Why

The `.duckdb` output is currently a copy of the CSV rendering: every column is `VARCHAR`, missing
values are the `-` placeholder, multi-value fields are packed into one delimiter-joined string, and
the file says nothing about what produced it. That is right for a spreadsheet, but `.duckdb` files
are read by BI tools, notebooks and generated SQL — all of which pay for the same workarounds:

| Problem | Cost |
|---|---|
| `-` (and `''`) used as NULL sentinels | `ErrorCode <> '-'` instead of `IS NOT NULL`; silently wrong results on 421,514 rows |
| Everything is `VARCHAR` | Casts in every view; a hand-written `CASE` rank for severity |
| 37% exact-duplicate rows | Every count and Top-N inflated ~1.6x, with no indication |
| `Tags` packs tactics and technique IDs together | `string_split` + `unnest` + a "starts with T + digits" heuristic |
| `"AWS-Region"` | Double-quoting in every statement, forever |
| No metadata table | A DFIR report cannot cite which ruleset produced a detection |

CSV/JSON output is unchanged — placeholders and packed strings belong there. Numbers measured against
`sample-aws-ct-timeline.duckdb` (1,925,150 rows).

## What changed

**1. Every file carries a `suzaku_meta` table** — `schema_version` (currently `1`), `suzaku_version`,
`command`, `command_line`, `generated_at`, `timestamp_tz` (`UTC` or the local offset under
`--localtime`; the values carry no offset, so this is otherwise unknowable), `rules_version` (the
rules folder's git HEAD) and `rules_count`, `scanned_files`/`scanned_events`, `output_rows` and
`duplicate_rows_removed`. Applies to `aws-ct-timeline`, `azure-timeline`, `aws-ct-search` and
`aws-ct-summary`.

**2. `NULL` instead of `-` / `''`** in DuckDB, so `IS NULL` is the right question,
`count(DISTINCT UserName)` is no longer off by one, and `-` stops appearing as a BI category.

**3. Real types** — `Timestamp TIMESTAMP`, `Level suzaku_level`
(`ENUM('informational','low','medium','high','critical')`), `Tactics`/`TechniqueIDs`/`OtherTags` as
`VARCHAR[]`, and typed summary timestamps. `Level` is a **named** type on purpose: DuckDB compares an
ENUM against a bare literal as text, so `Level >= 'high'` would mean the alphabetical
`'high' <= 'informational'`. With the named type the threshold filter works:

```sql
SELECT Timestamp, RuleTitle, TechniqueIDs FROM timeline
WHERE Level >= 'high'::suzaku_level
  AND Timestamp BETWEEN TIMESTAMP '2024-01-01' AND TIMESTAMP '2024-02-01'
  AND ErrorCode IS NULL;          -- the call succeeded
```

`ORDER BY Level DESC` and `max(Level)` now rank by severity without a `CASE`. Unparseable values
degrade to `NULL` (`TRY_CAST`) rather than failing the write.

**4. Exact-duplicate rows removed, and the count reported.** 719,101 of 1,925,150 rows (37.4%) were
duplicates — the same `(EventID, RuleID)` up to 8 times with all 20 columns equal. A row is dropped
when it is identical to another in *every* column after the normalization above, so two rows that
differed only in which NULL sentinel they used, or in an unparseable timestamp, also collapse. Such
rows carry no information. The legitimate multi-rule fan-out is kept — the grain stays
`event × rule match`, so an event matching several rules still produces one row per match and
`EventID` remains non-unique. Since the cause is usually upstream (overlapping CloudTrail exports),
the count is reported in `suzaku_meta.duplicate_rows_removed` rather than silently discarded.

**5. Multi-value and composite columns unpacked.** Timeline `Tags` becomes the three list columns
above, with `OtherTags` keeping group IDs, CVEs and any other namespace so nothing is lost. In the
summary: `Category` (`abused_success` / `other_failed`, which forced `LIKE 'abused%'` for one fact
and `LIKE '%failed'` for the other) becomes `IsAbused BOOLEAN` + `Outcome ENUM('success','failed')`;
`API` (`'RunInstances (ec2.amazonaws.com)'`) splits into `API` + `EventSource`, matching the timeline;
`UserTypes` is a `VARCHAR[]` that now lists *every* identity type a principal was seen with (the
aggregator assigned rather than accumulated, so only the last one survived).

**6. Consistent naming.** `"AWS-Region"` → `AwsRegion`, and `summary_attributes.Attribute` values are
spelled like the timeline columns holding the same fact
(`aws_region`/`src_ip`/`access_key_id`/`user_agent` → `AwsRegion`/`SrcIP`/`UserAccessKeyID`/`UserAgent`).

**7. Grain, sorting and durability.** The grain is recorded as a table comment readable from the file
(`SELECT comment FROM duckdb_tables()`) — `EventID` has 1,139,102 distinct values over 1,925,150 rows,
so assuming uniqueness is wrong 750,406 times. Rows are written `ORDER BY Timestamp` (236.2 → 201.7
MiB, −15%; no query-performance claim). `CHECKPOINT` before exit so the file has no leftover `.wal`
and opens from a read-only mount. Summary tables declare `NOT NULL` on the columns Suzaku generates
itself.

## Resulting schema

```sql
-- aws-ct-timeline / azure-timeline / aws-ct-search
CREATE TABLE timeline (
    Timestamp        TIMESTAMP,      -- was VARCHAR
    RuleTitle        VARCHAR,
    RuleAuthor       VARCHAR,
    Level            suzaku_level,   -- was VARCHAR
    EventName        VARCHAR,
    ErrorCode        VARCHAR,        -- NULL, not '-'
    ErrorMessage     VARCHAR,
    EventSource      VARCHAR,
    AwsRegion        VARCHAR,        -- was "AWS-Region"
    SrcIP            VARCHAR,
    SrcASN           VARCHAR,        -- present when --geo-ip is used
    SrcCity          VARCHAR,
    SrcCountry       VARCHAR,
    UserAgent        VARCHAR,
    UserName         VARCHAR,
    UserType         VARCHAR,
    UserAccountID    VARCHAR,
    UserARN          VARCHAR,
    UserPrincipalID  VARCHAR,
    UserAccessKeyID  VARCHAR,
    EventID          VARCHAR,
    Tactics          VARCHAR[],      -- was packed into Tags
    TechniqueIDs     VARCHAR[],
    OtherTags        VARCHAR[],
    RuleID           VARCHAR
);

-- aws-ct-summary
CREATE TABLE summary (
    UserARN         VARCHAR   NOT NULL,
    NumOfEvents     BIGINT    NOT NULL,
    FirstTimestamp  TIMESTAMP,
    LastTimestamp   TIMESTAMP,
    UserTypes       VARCHAR[]
);
CREATE TABLE summary_api_calls (
    UserARN      VARCHAR NOT NULL,
    IsAbused     BOOLEAN NOT NULL,                   -- was Category
    Outcome      ENUM('success','failed') NOT NULL,  -- was Category
    API          VARCHAR NOT NULL,                   -- action only
    EventSource  VARCHAR,                            -- split out of API
    Description  VARCHAR,
    Count        BIGINT  NOT NULL,
    FirstSeen    TIMESTAMP,
    LastSeen     TIMESTAMP
);
CREATE TABLE summary_attributes (
    UserARN    VARCHAR NOT NULL,
    Attribute  VARCHAR NOT NULL,   -- AwsRegion | SrcIP | UserAccessKeyID | UserAgent
    Value      VARCHAR,
    Count      BIGINT  NOT NULL,
    FirstSeen  TIMESTAMP,
    LastSeen   TIMESTAMP
);

-- aws-ct-metrics
CREATE TABLE metrics (
    Field           VARCHAR NOT NULL,   -- CloudTrail field path, exactly as given to -F ('eventName')
    TimelineColumn  VARCHAR,            -- the timeline column holding the same fact ('EventName')
    Value           VARCHAR,            -- NULL when the event carried no value for Field
    Count           BIGINT  NOT NULL,
    FieldTotal      BIGINT  NOT NULL,   -- denominator: events counted for this Field
    Percent         DOUBLE  NOT NULL,   -- Count / FieldTotal, full precision
    FirstSeen       TIMESTAMP,          -- was VARCHAR
    LastSeen        TIMESTAMP,          -- was VARCHAR
    SrcASN          VARCHAR,            -- these three only when --geo-ip is used
    SrcCity         VARCHAR,
    SrcCountry      VARCHAR
);
```

Plus `suzaku_meta` in every file.

## Implementation notes

* New `src/core/duckdb_out.rs` holds the conventions shared by the timeline sink and the summary
  writer (NULL/timestamp/list expressions, `suzaku_meta`, comments, checkpoint) so the two cannot
  drift apart.
* Rows still reach DuckDB through the `Appender` in 10,000-row batches — the hot path is unchanged —
  but land in a **`TEMP`** staging table converted by one `INSERT`/`CREATE TABLE AS` at the end.
  Typing in one pass is what lets a single unparseable timestamp degrade to `NULL` instead of
  aborting the run, and is the only way to produce `LIST` columns (the appender cannot bind them).
  Nothing raw ends up in the finished file; cost is one extra pass at the end of a run.
* `process_events_from_dir` now returns the number of files read, for `suzaku_meta.scanned_files`.

## Compatibility

**Breaking for readers of the previous DuckDB output** — which is why `schema_version` exists from
now on. CSV/JSON/JSONL are byte-for-byte unchanged, except that summary `UserTypes` can now hold more
than one value (a bug fix). Everything is bundled into one release because 2.0.0 is already breaking
for the `-t/--output-type` rework, so consumers migrate once.

One consequence worth stating: because deduplication applies to DuckDB only, the same run can now
produce a different row count in `.duckdb` than in `.csv`. The difference is exactly
`suzaku_meta.duplicate_rows_removed`.

## Not in this PR

* **`snake_case` column names** — far more breakage for far less value than the hyphen fix.
* **Primary keys** — after dedup `(EventID, RuleID)` is *nearly* unique but not guaranteed, and a
  wrong constraint fails the write. The grain is a table comment instead.
* **Provenance columns (`SourceFile`, `RecordIndex`)** — would additionally answer "which file did
  this come from?", and can be added later without changing what is here.
* **Translated docs** — `dfir-timeline.md` and `dfir-summary.md` are updated; other locales need a
  separate sync.

## Testing

`cargo test` — 159 passing, with new tests pinning each change: placeholders become `NULL`;
`Timestamp` is temporal and one unparseable value becomes `NULL` instead of failing the write;
`Level` orders by severity and `>= 'high'::suzaku_level` works after reopening; duplicates are
dropped and counted; `Tags` splits into three lists (empty lists for untagged rules); `suzaku_meta`
is single-row and the grain comment is present; the file has no pending WAL and opens read-only; the
summary's split columns, list `UserTypes` and timestamp types.

Verified end to end against `test_files/json` for all four commands, with `--localtime`, `--geo-ip`
and a re-run with `-C, --clobber` over an existing database.


YamatoSecurity:
@fukusuket Sorry, I think I caused some conflicts after merging other PRs. Please check.

fukusuket:
@YamatoSecurity 
Thank you for checking! The conflicts have been resolved.
Plus, I've pushed a commit with an update to the DuckDB schema in aws-ct-metrics :)

YamatoSecurity (approved):
@fukusuket LGTM! Thanks!!

PR #181, opened 2026-08-03
test: make the #159 GeoIP regression test able to fail

### LINKED ISSUE #159: Azure/M365 GeoIP enrichment is a silent no-op (source-IP field hardcoded to `sourceIPAddress`)

Shirofune-Security:
Follow-up to #171 (which fixed #159). **Test-only — no production code changes.**

## The problem

The regression test added alongside the #159 fix passed whether or not the fix was present. I reintroduced the exact bug it guards — resolving the source IP from a hardcoded `event.get("sourceIPAddress")` instead of the profile's `SrcIP` spec — and it stayed green.

Both reasons are about the fixture, not the assertions:

**The IP was not in the test databases.** The test looked up `8.8.8.8`. `test_files/mmdb/` holds MaxMind's small *test* fixtures — about 240 networks — not the real GeoLite2 feeds, and `8.8.8.8` is not among them. So the AWS control and the Azure subject both returned the `-` placeholder and `assert_eq!(aws_country, azure_country)` compared `"-"` with `"-"`.

**The one assertion that could have caught it was guarded.**

```rust
assert_eq!(aws_country, azure_country);   // "-" == "-", passes with the bug
if aws_country != "-" {                   // never true with these fixtures
    assert_ne!(azure_country, "-");
}
```

## What changed

The lookups use `89.160.20.112`, which the checked-in fixtures resolve in all three databases (ASN `Bredband2 AB`, city `Linköping`, country `Sweden`). The guard is gone and the assertions name concrete values.

The AWS lookup stays as an explicit control carrying a message that points at the fixture:

```rust
assert_eq!(
    lookup("SrcCountry", &aws_event, ".sourceIPAddress", &mut geo),
    "Sweden",
    "the control lookup must enrich; is 89.160.20.112 still in test_files/mmdb?"
);
```

so if the databases are ever swapped the test says *why* it broke instead of quietly going vacuous again.

Coverage also widened past the single `SrcCountry` lookup:

- all three geo columns, since they are three separate lookups against three separate databases;
- a `|`-separated spec where the field actually present is **not** the first candidate (the real Azure/M365 profile is `.claims.ipaddr|.callerIpAddress|.ClientIP|.ActorIpAddress`, and Azure activity logs frequently carry `callerIpAddress` without `claims.ipaddr`);
- an empty spec — a profile with no `SrcIP` column;
- an event carrying none of the spec's candidates, which must yield `-` rather than enriching from an unrelated field.

## Verification

Mutation-tested, each run against the previous test and this one:

| mutation | before | after |
|---|---|---|
| hardcode `sourceIPAddress` (the #159 bug) | passed | **fails** — `left: "-", right: "Sweden"` |
| take only the first candidate, don't walk the `\|` list | passed | **fails** |
| ignore the profile spec, use a fixed candidate list | passed | **fails** |

186 tests pass; `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

## On #159 itself

The production fix from #171 is correct and I confirmed it end to end rather than from the code — `azure-timeline` with `callerIpAddress` `89.160.20.112` and `-G test_files/mmdb` enriches to `Bredband2 AB` / `Linköping` / `Sweden`, where `main` before #171 produced `-`. **#159 can be closed**; it appears to have stayed open only because #171 did not reference it in a closing keyword.

## Noticed while testing, filed separately

`2001:218::1` and `240d:1a:7fe:a000:…` are reported as `SrcASN: Private`. Both are global unicast (`2000::/3`), not private. hayabusa had the same bug and fixed it by removing `2000::/3` and `FD00::/8` from its private-range list. Not touched here.

### DIFF
--- a/src/core/timeline_writer.rs
+++ b/src/core/timeline_writer.rs
@@ -1455,6 +1455,202 @@ mod tests {
         );
     }
 
+    /// Drives the real `write_record` path -- the shipped profile from `load_profile`, the
+    /// profile-derived `SrcIP` spec, the CSV writer -- and returns the emitted row as a map.
+    ///
+    /// The point is to bind the profile to the enrichment. `geoip_resolves_source_ip_via_profile_spec`
+    /// passes literal specs straight to `get_value_from_event`, so hardcoding a field name back
+    /// into `write_record` (which is where #159 actually lived) would leave every assertion there
+    /// green. Verified by mutation: replacing `src_ip_spec(context.profile)` at the top of
+    /// `write_record` with a literal `".sourceIPAddress"` fails this test and nothing else in the
+    /// suite.
+    fn write_record_columns(
+        log: &crate::core::log_source::LogSource,
+        event_json: &str,
+        geo: &mut Option<crate::option::geoip::GeoIPSearch>,
+    ) -> std::collections::HashMap<String, String> {
+        use crate::core::util::load_profile;
+        use sigma_rust::{event_from_json, rule_from_yaml};
+
+        let profile = load_profile(log, geo, true);
+        let rule = rule_from_yaml(
+            "title: t\nlogsource:\n    category: test\ndetection:\n    selection:\n        eventName: E\n    condition: selection\n",
+        )
+        .unwrap();
+        let event = event_from_json(event_json).unwrap();
+        let json: Value = serde_json::from_str(event_json).unwrap();
+
+        let dir = tempfile::tempdir().unwrap();
+        let csv_path = dir.path().join("out.csv");
+        let config = OutputConfig::new(true, false, false);
+        {
+            let file = std::fs::File::create(&csv_path).unwrap();
+            let writers = Writers::new()
+                .with_csv(csv::WriterBuilder::new().from_writer(Box::new(file) as Box<dyn Write>));
+            let mut context = OutputContext::new(&profile, geo, &config, writers, &[]);
+            write_record(&event, &json, Some(&rule), &mut context);
+            context.flush_all();
+        }
+
+        let text = std::fs::read_to_string(&csv_path).unwrap();
+        let mut reader = csv::ReaderBuilder::new()
+            .has_headers(false)
+            .from_reader(text.as_bytes());
+        let row: Vec<String> = reader
+            .records()
+            .next()
+            .expect("write_record emitted no CSV row")
+            .unwrap()
+            .iter()
+            .map(|f| f.to_string())
+            .collect();
+        profile.iter().map(|(k, _)| k.clone()).zip(row).collect()
+    }
+
+    // #159 must stay fixed at the level it actually broke: the profile. The test above pins the
+    // resolution given a spec; this one pins that the shipped Azure/M365 profile is what supplies
+    // that spec, end to end through `write_record`.
+    #[test]
+    fn write_record_enriches_geoip_from_the_shipped_profile() {
+        use crate::core::log_source::LogSource;
+        use crate::option::geoip::GeoIPSearch;
+        use std::path::Path;
+
+        let mut geo = Some(
+            GeoIPSearch::new(Path::new("test_files/mmdb"))
+                .expect("GeoLite2 test .mmdb files must be present under test_files/mmdb/"),
+        );
+
+        // Azure: the caller IP lives in `callerIpAddress`, a field name that does not exist in
+        // CloudTrail. This is the exact record shape #159 reported as always rendering "-".
+        let azure = write_record_columns(
+            &LogSource::Azure,
+            r#"{"callerIpAddress": "89.160.20.112", "eventName": "E", "operationName": "op"}"#,
+            &mut geo,
+        );
+        assert_eq!(
+            azure.get("SrcIP").map(String::as_str),
+            Some("89.160.20.112")
+        );
+        assert_eq!(
+            azure.get("SrcASN").map(String::as_str),
+            Some("Bredband2 AB")
+        );
+        assert_eq!(azure.get("SrcCity").map(String::as_str), Some("Linköping"));
+        assert_eq!(azure.get("SrcCountry").map(String::as_str), Some("Sweden"));
+
+        // AWS through the same path, so a regression that breaks one log source and not the other
+        // is still caught.
+        let aws = write_record_columns(
+            &LogSource::Aws,
+            r#"{"sourceIPAddress": "89.160.20.112", "eventName": "E"}"#,
+            &mut geo,
+        );
+        assert_eq!(aws.get("SrcASN").map(String::as_str), Some("Bredband2 AB"));
+        assert_eq!(aws.get("SrcCountry").map(String::as_str), Some("Sweden"));
+    }
+
+    /// The correlation sibling of `write_record_columns`: drives `write_correlation_record`,
+    /// which builds its row through `get_value_from_correlation_event` and computes the spec at a
+    /// *different* call site (`build_correlation_record`). A hardcode there would be invisible to
+    /// the non-correlation tests.
+    fn write_correlation_columns(
+        log: &crate::core::log_source::LogSource,
+        event_json: &str,
+        geo: &mut Option<crate::option::geoip::GeoIPSearch>,
+    ) -> std::collections::HashMap<String, String> {
+        use crate::core::util::load_profile;
+        use sigma_rust::{SigmaCorrelationRule, TimestampedEvent, event_from_json, rule_from_yaml};
+
+        let profile = load_profile(log, geo, true);
+        let base_rule = rule_from_yaml(
+            "title: t\nlogsource:\n    category: test\ndetection:\n    selection:\n        eventName: E\n    condition: selection\n",
+        )
+        .unwrap();
+        let correlation_rule = SigmaCorrelationRule {
+            title: "c".to_string(),
+            ..Default::default()
+        };
+        let timestamped = TimestampedEvent {
+            event: event_from_json(event_json).unwrap(),
+            timestamp: "2024-01-01T00:00:00Z".parse().unwrap(),
+            rule: &base_rule,
+        };
+
+        let dir = tempfile::tempdir().unwrap();
+        let csv_path = dir.path().join("out.csv");
+        let config = OutputConfig::new(true, false, false);
+        {
+            let file = std::fs::File::create(&csv_path).unwrap();
+            let writers = Writers::new()
+                .with_csv(csv::WriterBuilder::new().from_writer(Box::new(file) as Box<dyn Write>));
+            let mut context = OutputContext::new(&profile, geo, &config, writers, &[]);
+            write_correlation_record(&vec![&timestamped], &correlation_rule, &mut context);
+            context.flush_all();
+        }
+
+        let text = std::fs::read_to_string(&csv_path).unwrap();
+        let mut reader = csv::ReaderBuilder::new()
+            .has_headers(false)
+            .from_reader(text.as_bytes());
+        let row: Vec<String> = reader
+            .records()
+            .next()
+            .expect("write_correlation_record emitted no CSV row")
+            .unwrap()
+            .iter()
+            .map(|f| f.to_string())
+            .collect();
+        profile.iter().map(|(k, _)| k.clone()).zip(row).collect()
+    }
+
+    // Correlation rows go through their own record builder and their own `src_ip_spec` call, so
+    // #159 could regress there alone. Verified by mutation: hardcoding the spec inside
+    // `build_correlation_record` fails this test and nothing else.
+    #[test]
+    fn write_correlation_record_enriches_geoip_from_the_shipped_profile() {
+        use crate::option::geoip::GeoIPSearch;
+        use std::path::Path;
+
+        let mut geo = Some(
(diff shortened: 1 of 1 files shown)


Shirofune-Security:
Both confirmed — thanks, these were right and the first one was a real hole in what I'd added. Fixed in 0a6eaf0; the second is production behaviour so I've split it out.

## P2 — test bypasses the production profile wiring: fixed

Confirmed by mutation before fixing: replacing `src_ip_spec(context.profile)` in `write_record` with a literal `".sourceIPAddress"` — #159 recreated at the caller — left **all 186 tests green**.

Three tests now drive the real path (`load_profile` for the shipped profile → `write_record` / `write_correlation_record` → emitted CSV row):

- `write_record_enriches_geoip_from_the_shipped_profile` — Azure `callerIpAddress` and AWS `sourceIPAddress` through the same path, so a regression that breaks one log source and not the other is still caught.
- `write_correlation_record_enriches_geoip_from_the_shipped_profile` — you were right to ask for this separately: `build_correlation_record` computes the spec at its **own** call site, so a hardcode there is invisible to the record path.
- `write_record_omits_geo_columns_without_geoip` — the negative, so the enrichment above cannot be an artifact of columns that are always present.

Each mutation now fails only its own test:

| mutation | before | after |
|---|---|---|
| hardcode the spec in `write_record` | 186 passed | **fails** `write_record_enriches_…` |
| hardcode the spec in `build_correlation_record` | 186 passed | **fails** `write_correlation_record_enriches_…` |

## P2 — fallback only covers an absent first candidate: confirmed, filed as #183

Reproduced end to end. Every row below has a valid `callerIpAddress` of `89.160.20.112`:

| `claims.ipaddr` | `SrcIP` | `SrcCountry` |
|---|---|---|
| `""` / `"-"` / `"not-an-ip"` / `"89.160.20.112:443"` | the bad value | `-` |
| *absent* | `89.160.20.112` | **Sweden** |

I've left this out of #181 and filed it as **#183** rather than folding it in, because the fix changes production behaviour while this PR is test-only — and because there are two design calls in it that deserve their own review rather than riding along on a test PR:

1. If only the geo lookup switches to first-*parseable*, a record can show `SrcIP: 89.160.20.112:443` next to `SrcCountry: Sweden` resolved from a different field — the `SrcIP` column is rendered by the generic `.a|.b` resolver, which has the same first-present semantics. Making both consistent is a wider change.
2. `IP:port` is the single most likely real-world trigger (M365 UAL `ClientIP` is routinely `1.2.3.4:56789`), and stripping the port would *enrich* those rather than fall through them — arguably a better fix than skipping, and a separate enhancement.

#183 carries the repro, the suggested fix, and a note that whatever lands should add null/empty/placeholder/non-parseable first-candidate cases.

YamatoSecurity (approved):

PR #193, opened 2026-09-01
fix: preserve GeoIP enrichment in raw output

### LINKED ISSUE #184: --raw-output silently discards GeoIP enrichment (-G becomes a no-op)
## Summary

With `--raw-output`, `-G/--geo-ip` is silently a no-op: the `SrcASN`/`SrcCity`/`SrcCountry` columns are absent from the emitted JSON/JSONL entirely. No warning is printed, and the flag is accepted.

This is not "raw means raw" — the same code path deliberately augments the raw record with five derived columns; geo is simply not among them.

## Reproducing

Same input, same `-G`, one flag apart:

```bash
suzaku azure-timeline -f az.json -G test_files/mmdb -o /tmp/norm.jsonl  -t jsonl -C
suzaku azure-timeline -f az.json -G test_files/mmdb -o /tmp/raw.jsonl   -t jsonl --raw-output -C
```

```
norm.jsonl  {"SrcASN": "Bredband2 AB", "SrcCity": "Linköping", "SrcCountry": "Sweden", "SrcIP": "89.160.20.112", ...}
raw.jsonl   (no Src* keys at all)
```

The raw record *does* carry the sigma-derived columns: `RuleTitle`, `Level`, `RuleAuthor`, `Tags`, `RuleID`.

## Cause

`src/core/timeline_writer.rs:502-537`. In raw mode the record is `json.clone()` plus the profile entries whose value starts with `"sigma."`:

```rust
let sigma_profile: Vec<(String, String)> = profile
    .iter()
    .filter(|(_, value)| value.starts_with("sigma."))
    .cloned()
    .collect();
```

`load_profile` injects the geo columns as `("SrcASN", "SrcASN")`, `("SrcCity", "SrcCity")`, `("SrcCountry", "SrcCountry")` (`src/core/util.rs:204-208`) — the values are bare column names, not `sigma.`-prefixed, so the filter drops them. The enrichment is computed and discarded.

## Why it matters

`--raw-output` exists to feed the original logs into another system, which is exactly the workflow where the geo columns are most useful — the downstream consumer cannot recompute them without its own MaxMind database. An analyst who passes `-G --raw-output` gets no enrichment and no indication that the flag did nothing. Same silent-no-op failure mode as #159.

## Suggested fix

Either:

- **enrich** — treat the geo columns like the sigma ones and add them to the raw record (consistent with the mode already augmenting; the geo values are computed on this path anyway); or
- **refuse** — reject `-G` together with `--raw-output` with a clear message, the way other incompatible flag pairs are handled.

Enriching seems the better fit given the mode already adds five derived columns, but either beats accepting the flag and ignoring it.

## Coverage note

No test exercises `--raw-output` with `-G`, which is why this was invisible.

vitorparras:
## What Changed

- Preserve GeoIP-derived fields in `--raw-output` while retaining the original event payload and Sigma fields.
- Use one enrichment path for terminal output, JSON, and JSONL so their behavior cannot diverge.
- Add regression coverage using the repository's GeoLite2 fixtures for both JSON and JSONL output.

Closes #184

## Evidence

- `cargo fmt -- --check` passed.
- `cargo test --verbose` passed: 198 tests.
- `cargo clippy --all-targets -- -D warnings` passed.
- Ran the real CLI in Docker with `azure-timeline`, `--raw-output`, `-G test_files/mmdb`, and JSON output. The generated record retained `callerIpAddress`, `RuleTitle`, `SrcASN`, `SrcCity`, and `SrcCountry`.

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -4,6 +4,7 @@
 
 **バグ修正:**
 
+- `-G, --geo-ip` で一致したタイムラインイベントを GeoIP エンリッチメントしていても、`--raw-output` は `SrcASN`・`SrcCity`・`SrcCountry` を出力せず、GeoIP が無効であるかのように見えていた。生のターミナル出力、JSON、JSONL が同じプロファイル由来のエンリッチメント経路を使用するようにし、元のイベントフィールドと Sigma メタデータを保持したまま GeoIP フィールドを追加するようにした。 (#184) (@vitorparras)
 - イベントのタイムスタンプ取得時に、出力プロファイルのフィールド指定をそのまま `Event::get` に渡していた問題を修正した。プロファイルの指定は先頭にドットが付き `|` 区切りのフォールバックリストにもなる（`.eventTime`、`.time|.eventTimestamp|.CreationTime`）一方、`Event::get` はドットなしのフィールド名を受け取るため、常に `None` が返っていた。この影響で、サマリーの `Dates with most total detections` が `n/a` となり、`first_event_time`/`last_event_time` が設定されず、さらに **Sigma の相関（correlation）ルールが一切発火しなかった**（同梱の相関ルール5本がすべて無効）。該当する3箇所すべてで、`Timestamp` カラムと同じ「存在する最初の候補を採用する」ルールの共通ヘルパー経由で解決するようにし、表示値と分析値が食い違わないようにした。 (#191) (@fukusuket)
 
 **その他:**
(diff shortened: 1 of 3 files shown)


YamatoSecurity:
@vitorparras Thanks for this! Can you also update the two changelogs? 🙏

vitorparras:
> @vitorparras Thanks for this! Can you also update the two changelogs? 🙏

Done — I have updated both `CHANGELOG.md` and `CHANGELOG-Japanese.md` with the requested entry for #184.

fukusuket (approved):
@vitorparras Thank you! I confirmed that the -G option works with --raw-output. LGTM!!🚀

PR #194, opened 2026-09-14
feat: gws-timeline — Google Workspace Reports API support

ecapuano:
Adds Google Workspace as a third log source alongside CloudTrail and Azure/M365, following
the pattern `azure-timeline` established in v2.0: `LogSource::Gws`, `config/gws_profile.yaml`,
a `normalize_gws_event()` step in `scan.rs`, service matching for SigmaHQ's
`google_workspace.*` logsources, and a `gws-timeline` subcommand. English docs, fixtures and
tests included. Companion rules PR in suzaku-rules adds `suzaku/gws/` (18 detection rules,
2 correlation rules) and `config/gws_ignore_rule_list.txt`.

## Input

Google Admin SDK Reports API `activities.list` items (`"kind": "admin#reports#activity"`) as
`.json`, `.jsonl` or `.json.gz`, in any of the shapes the API and the exporters around it
produce:

- a bare activity object, or a JSON array of activities
- an `activities.list` response page (`{"kind":"admin#reports#activities","items":[...]}`)
- JSONL with one of the above per line

The Admin console CSV export is a different, lossy schema and is out of scope here.

## Why normalization is required

Each activity carries an `events[]` array of 1..N sub-events, each with its own
`parameters[]` name/value list. Sigma rules cannot address array elements, and the
interesting semantics often live past the first element: an admin `CHANGE_PASSWORD` arrives
together with `CHANGE_PASSWORD_ON_NEXT_LOGIN`; a Calendar `create_event` carries one
`add_event_guest` sub-event per guest (a 500-guest phishing invite is one API record).
`normalize_gws_event()` therefore emits one record per `events[i]`:

- copies every top-level activity field (`kind`, `id`, `actor`, `ipAddress`, `networkInfo`,
  `resourceDetails`, `ownerDomain`)
- adds `eventName`, `eventType`, `eventService = "<applicationName>.googleapis.com"` (only
  when `id.applicationName` is present), `eventIndex`, `eventCount`, and the sub-event's
  `resourceIds`; Suzaku-generated keys always win over a same-named parameter
- caps splitting at 10,000 sub-events per activity with a warning, since each record copies
  the activity's top-level fields and an adversarial input could otherwise amplify a few KB
  of gzip into gigabytes; `eventCount` still reports the real length
- an activity with an empty or missing `events[]` is emitted as-is rather than dropped
- folds `parameters[]` into top-level keys by name (`value` / `intValue` parsed to i64 /
  `boolValue` / `multiValue` / `multiIntValue` / `messageValue` folded recursively /
  `multiMessageValue`)
- drops a parameter that carries no value at all: the API elides protobuf defaults, so a
  bare `{"name":"is_suspicious"}` means false (in a multi-tenant sample it is emitted on
  every `login_success`, valueless about 13x more often than as `boolValue: true`), and
  folding it to `true` would make `is_suspicious: true` fire on every login
- adds a lowercase alias for every `UPPER_CASE` parameter name (`NEW_VALUE` -> `new_value`)
  without ever overwriting a natively lowercase key, so SigmaHQ's gworkspace admin rules
  (Elastic Filebeat field names) load unmodified
- does NOT synthesize Cloud Logging `protoPayload.*` fields. SigmaHQ's three
  `gworkspace/login` rules are written against that schema and will not load here; we chose
  not to inject synthetic fields into every output format and will propose converging those
  rules on the Reports API schema upstream instead (see question 2)

Same idea as the M365 `ExtendedProperties` folding already in `scan.rs`.

## Service matching

`supported_services()` for Gws: `google_workspace` plus `google_workspace.<app>` for admin,
login, drive, calendar, token, user_accounts, saml, groups, mobile, gmail, chat, meet, chrome,
rules, context_aware_access, access_transparency, keep, vault. `is_match_service()` requires
`kind == "admin#reports#activity"` and, for the dotted form, `id.applicationName == <app>`.

## Output profile

`config/gws_profile.yaml` (23 columns). Google already attaches `networkInfo.ipAsn` and
`regionCode`, exposed as `IpASN` / `IpRegion`; `-G` still appends `SrcASN`/`SrcCity`/`SrcCountry`
from MaxMind as for the other sources. The profile keys deliberately avoid those three names so
the header never carries duplicates. `Timestamp` is the nested `id.time`; `filter_by_time` gained
a nested-path lookup so `--timeline-start` / `--time-offset` work on it (flat keys unchanged).

## Fixes to existing code made along the way

- `write_correlation_record` never set `has_written`, so a run where only correlation rules
  fired printed "Results saved: None" and deleted its own output files. Fixed, with a
  mutation-verified regression test.
- In the dependency, sigma-rust 0.7.1 deserializes `group_by` while the Sigma spec and every
  shipped correlation rule spell it `group-by`. Serde ignores the unknown key, so grouping is
  silently dropped and thresholds are evaluated corpus-wide; seven of the AWS CloudTrail
  correlation rules in suzaku-rules are affected today. A separate sigma-rust PR adds
  `#[serde(rename = "group-by", alias = "group_by")]` with tests. This PR does not depend on
  it, but correlation rules will not group correctly until it lands.

## Testing

- `cargo test` 217 passed (21 new: splitting, folding, key precedence, aliasing, sub-event
  cap, empty `events[]`, response-page unwrap, service matching, profile
  timestamp resolution, nested time filter, correlation output retention); `cargo clippy
  --all-targets -- -D warnings` and `cargo fmt --check` clean.
- End-to-end on a real multi-tenant corpus of 197,013 activities (298,867 records after the
  split) with the companion rules: 3.1 s, 1.2 GB peak RSS, all 18 detection rules matched an
  independent Python reference count exactly (the two noisiest rules were tuned from 737
  and 682 weekly hits to 55 and 36 by excluding Google Ads re-authentication prompts), error log never created; gzip input, time
  filtering, DuckDB output (`suzaku_meta.command = gws-timeline`) verified.

## Test data

`test_files/json/gws/` ships two small fixtures: 151 login/admin activities as JSONL and a
20-activity Calendar response page. They come from a deny-by-default pseudonymization pass over
real tenant data: every string leaf is either a Google enum, a deterministic hash or
placeholder, a pseudonymous `example-N.test` address, or a TEST-NET address, verified by an
automated leaf-level check and by intersection against the source corpus. `.gitignore` gained `!test_files/` negations so the fixtures are tracked normally instead of
needing `git add -f` like the existing `test_files/json/*`.
Token, user_accounts, saml, groups and gmail samples will follow once our collection scope
widens.

## Questions for maintainers

1. Command name: `gws-timeline` as implemented, or `google-workspace-timeline`?
2. SigmaHQ's `gworkspace/login` rules use Cloud Logging `protoPayload.*` paths and therefore do
   not load against Reports API records. We would rather converge them upstream than emit
   synthetic fields; objections?
3. `--raw-output` emits the split record. Should it emit the original activity instead?
4. Any objection to a later PR accepting the Admin console CSV export? IR teams are handed
   that format far more often than API JSON.

## Related

- Rules: https://github.com/Yamato-Security/suzaku-rules/pull/83
- sigma-rust group-by fix: https://github.com/Yamato-Security/sigma-rust/pull/8


ecapuano:
Tested all ten SigmaHQ `rules/cloud/gcp/gworkspace` rules as-is against this branch, since the same exercise turned up problems for Azure.

**Result**

| Rules | Load | Match Reports API records |
|---|---|---|
| 7 admin (`eventService`, `eventName`, `setting_name`, `new_value`) | yes | yes, every event name they reference |
| 3 login (`protoPayload.serviceName`, `protoPayload.metadata.event.eventName`) | yes | no, by construction |

The admin rules use the flat Reports API field style plus lowercase parameter names, which the `UPPER_CASE` aliasing here satisfies. Both string `"false"` and `boolValue: false` satisfy the MFA rule's `new_value: 'false'`.

The login rules were written against the Cloud Logging export of Workspace audit logs (`protoPayload.*`), a different schema from the Reports API, so they can never match here. They are otherwise valid for that schema: with Google's published Cloud Logging login samples, two of the three match once the `metadata.event` array is flattened. The third, `Suspicious Login Activity Classified By Google`, has a casing typo (`protoPayload.Servicename`) and matches nothing anywhere until it is corrected; fix submitted upstream. The companion `suzaku/gws/login` rules cover the same five event names on the Reports API schema.

**Reproduce** (fixture added in the latest commit, 20 synthetic activities incl. negative cases):

```
suzaku gws-timeline -f test_files/json/gws/sigmahq_rules_sample.jsonl -r <sigma>/rules/cloud/gcp/gworkspace
```

Expected: `Total detection rules: 10`, 13 hits, all from the seven admin rules.

Also in that commit: `Target` now falls back to `APPLICATION_NAME`, `API_CLIENT_NAME` and `SETTING_NAME`, so the events those admin rules match show what changed instead of `-`.

The suzaku-rules PR gained a commit that vendors `rules/cloud/gcp/gworkspace` in the sync pipeline (only that subtree; `gcp/audit` is Cloud Audit Logs, which Suzaku does not read) and lists the three login rule IDs in `filtered_sigma_rules.txt` with the reason above.

YamatoSecurity:
@ecapuano When you get a chance, can you check/fix/resolve all of the copilot comments?

fukusuket (approved):
Thank you so much for adding the new feature! LGTM!!🚀

YamatoSecurity:
@ecapuano 
> 1. Command name: gws-timeline as implemented, or google-workspace-timeline?
`gws-timeline` is good.

> 2. SigmaHQ's gworkspace/login rules use Cloud Logging protoPayload.* paths and therefore do
not load against Reports API records. We would rather converge them upstream than emit synthetic fields; objections?

Yes, we should probably fix upstream. It could be also good to create a config file that had the mapping of the field names used in Sigma rules and the alias field names used in the different logs and then Suzaku in the background will convert the field names to support the various log formats. Ex: `eventService` will also check `protoPayload.serviceName` etc..


> 3. --raw-output emits the split record. Should it emit the original activity instead?

split record sounds good.

> 4. Any objection to a later PR accepting the Admin console CSV export? IR teams are handed
that format far more often than API JSON.

Very welcome!

I'll go ahead and merge this for now so please send in any PRs to improve on it.

